PR #22, opened 2016-08-27
Changes manual build instructions

nskins:
While running through the manual build instructions, there were a few changes to the commands that were necessary to successfully build. Here is the summary:

- Specify 'sandbox' when adding the extra dependency,
- Specify the dependency's root directory rather than its .cabal file,
- Install the dependencies or else 'cabal build' will fail and report missing dependencies.


dagit:
Thanks!

PR #79, opened 2016-09-08
use Text for edit widgets

dagit:
Submitting this as a PR because it depends on my recent PRs to both text-zipper and brick. If those get merged then you'll probably want to merge this as well.


jtdaugherty:
brick 0.9 is out, so this will have to also increase our bound on brick, too.

PR #100, opened 2016-10-07
Clean up some import warnings

travitch:
Strategically import Prelude to avoid warnings about Control.Applicative and
Data.Monoid under ghc 7.10.

### DIFF
--- a/src/Command.hs
+++ b/src/Command.hs
@@ -8,6 +8,8 @@ import Control.Monad.IO.Class (liftIO)
 import Data.Monoid ((<>))
 import qualified Data.Text as T
 
+import Prelude
+
 import State
 import Types
 
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -18,6 +18,8 @@ import           Data.Monoid ((<>))
 import           System.Process (readProcess)
 import           Text.Read (readMaybe)
 
+import           Prelude
+
 import           IOUtil
 import           FilePaths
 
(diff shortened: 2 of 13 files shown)


dagit:
I think the current plan is to fix this with base-compat, in fact we have a
branch that does it but that work is blocked by two vty bugs that make it
so the ghc 7.8 build is failing on Travis.

>

jtdaugherty:
@dagit I wasn't aware that anyone knew what the bugs were, or that they were indeed Vty bugs. As the only active Vty maintainer I'd be interested to know what they are if you've narrowed anything down; can you open some tickets on the Vty project?

dagit:
The two bugs I know of are:

  * Latest (or just recent?) vty segfaults on resize, that causes us to stick to an older version, 5.8 (Since you helped to discover that bug I just assumed it's in the vty issue tracker)
  * vty 5.8 (and possibly others) triggers a template Haskell bug in ghc 7.8.4

The latter error looks like this (from https://travis-ci.org/aisamanra/matterhorn/jobs/165342128):

```
ghc: internal error: stg_ap_pp_ret
    (GHC version 7.8.4 for x86_64_unknown_linux)
    Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
Failed to install vty-5.8
```

I haven't been able to reproduce the crash locally. Eric recommended trying a version of vty 5.8 that uses lens (or even a newer microlens-th) instead of the microlens-th it currently uses to see if the bug is specific to the version of microlens-th that vty 5.8 uses.

I hope that helps.

dagit:
I just did a quick test of removing the upper bound on vty.  That allowed me to test vty-5.11 on travis, which has the same build failure.

dagit:
It looks like it's not realistic to support 7.8.4 so we are merging this. Thanks for the PR!

PR #208, opened 2017-04-05
Add a handy command for running matterhorn when built with new-build.

acw:
Just a handy script, to go along with new-install.sh and friends.


dagit:
Thanks!

PR #229, opened 2017-05-08
Relax constraint on containers to allow 0.5.7 or greater.

kquick:
A couple of small tweaks lose a little laziness in the implementation but adds support for the older containers library.  This is useful because containers is closely related to ghc versioning and is a basis for a large number of other libraries, so if the native environment doesn't yet support the containers 0.5.8 (released Aug 31, 2016) it is likely to have containers 0.5.7 (released Dec 17, 2015).

### DIFF
--- a/matterhorn.cabal
+++ b/matterhorn.cabal
@@ -63,7 +63,7 @@ executable matterhorn
                      , mattermost-api >= 30701.1.0
                      , base-compat
                      , unordered-containers
-                     , containers >= 0.5.8
+                     , containers >= 0.5.7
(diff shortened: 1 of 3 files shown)


jtdaugherty:
Can the check this patch be factored out as a new function? E.g. `safeIndex :: Seq a -> Int -> Maybe a`.

kquick:
Yes it can. I would prefer to defer that effort based on some other refactoring if that is OK.

jtdaugherty:
Okay, thanks.

PR #300, opened 2017-07-28
List CHANGELOG.md in extra-doc-files

RyanGlScott:
This way, the changelog will be linked to from Hackage, which is handy.


jtdaugherty:
Thanks!

Issue #310, opened 2017-09-01
Use different message layout based on the message characteristics


kellymclaughlin:
Currently all messages use the same layout, but for messages that are block elements or multi-line messages this can result in messages being heavily indented especially when they are generated by web hooks where the display name length is longer. I think this makes such messages harder to read, but this could be a subjective thing so I wanted to open this issue to see what people think about using a vertical layout for block type and multi-line messages in a more similar way to how the MM web client lays these messages out.

 I've created a [branch](https://github.com/matterhorn-chat/matterhorn/compare/master...kellymclaughlin:feature/rework-message-layout) on my fork where I've implemented this and I have some screenshots to illustrate what I mean.

Current matterhorn: 

![2017-08-31-19075e6e-c138-4bba-b832-3228e098892f](https://user-images.githubusercontent.com/29337/29990742-cb6f6c96-8f3a-11e7-9227-84ee4255710f.png)

My branch: 

![2017-08-31-b29f9122-fd17-4dcb-b54a-3f18887e556f](https://user-images.githubusercontent.com/29337/29990744-d09d45f8-8f3a-11e7-8f87-90c1b74270c3.png)

jtdaugherty:
I'm in favor of this change, provided there is still some indentation (say, two spaces' worth) to distinguish the message content from the author name. @kquick / @aisamanra - thoughts?

kquick:
I like it (with the suggested addition of small indentation).  

Also, thank you for such a great issue submission!  The justification along with actual before/after examples is really nice.

jtdaugherty:
@kquick I want to second that - the screenshots and branch really helped make this easy to evaluate!

jtdaugherty:
Unless @aisamanra has any objections, I'd like to see a pull request for this change.

kellymclaughlin:
I'll add the extra indentation and then get PR up for the change shortly. Thanks!

aisamanra:
I am 100% in support of this change!

PR #314, opened 2017-09-06
Use message characteristics to determine layout

kellymclaughlin:
Currently the message layout is the same for all messages. For block
messages or messages that span multiple lines this means the entire
message is indented after the end of the user's name. This commit
changes the message layout so that block type messages or messages
spanning multiple lines are laid out vertically so the message appears
under the user's name rather using a horizontal lay out. The
horizontal layout is still used for single line messages.

This PR is in response to issue #310.

The following screen shots illustrate how different messages are currently 
laid out and how they appear with this change for comparison:

Current matterhorn:

current

This branch:

2017-09-06-f9bfafca-abe2-468c-b4eb-31f972267732

### DIFF
--- a/src/Markdown.hs
+++ b/src/Markdown.hs
@@ -1,6 +1,7 @@
-{-# LANGUAGE ViewPatterns #-}
-{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE MultiWayIf       #-}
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thank you!

Issue #331, opened 2017-11-29
Select the previously selected channel on restart


abhin4v:
When the matterhorn client restarts (quit and start again), it always selects the "town-square" channel. The mattermost web client remembers the last selected channel for the user and selects that channel on restarts (close the tab, reopen it). 

This feature should be implemented for feature parity and for user convenience.

abhin4v:
FYI, I'm willing to work on this.

jtdaugherty:
Thanks, that's great! There are some moving parts to this that we need to get nailed down, such as:

* How will the last channel state be stored? My preference: use the XDG data directory for Matterhorn to store a data file containing "last run state" that includes the channel ID of the last-visited channel.

* When a user runs Matterhorn but no such state file is present, Town Square is selected.

* When the user is not a member of the last-viewed channel, Town Square is selected.

* When the last-viewed channel has been archived, Town Square is selected.

* Matterhorn writes the last run state to disk on clean shutdown only.

jtdaugherty:
This is now fixed in `develop`.

PR #337, opened 2017-12-06
Add functionality to restore last selected channel on program restart

abhin4v:
- The last selected channel is written to a file on program exit.
- This file is specific to a team so that last selected channel can be saved for different teams in the same directory.
- This file contains server and user info as well to match against on program start.
- On program start, this file is looked up for the current team, and if the file exists and contains valid data for the current server and user, the current channel is set to the channel found in the file.
- If any of the above conditions fail, the current channel is set to Town Square.
- Fixes issue #331.


src/RunState.hs
```diff
@@ -0,0 +1,78 @@
...
+import Types
+import Zipper (focusL)
+
+-- | Run state of the program. This is saved in a file on program exit and
+-- | looked up from the file on program startup.
+data RunState = RunState
+  { _rsHost      :: Hostname  -- ^ Host of the server
+  , _rsPort      :: Port      -- ^ Post of the server
+  , _rsUserId    :: UserId    -- ^ ID of the logged-in user
+  , _rsChannelId :: ChannelId -- ^ ID of the last selected channel
```
jtdaugherty: Can you rename this to `_rsLastSelectedChannelId`?

src/RunState.hs
```diff
@@ -0,0 +1,78 @@
...
+import FilePaths
+import Network.Mattermost (Hostname, Port)
+import Network.Mattermost.Types
+import Network.Mattermost.Lenses
+import Types
+import Zipper (focusL)
+
+-- | Run state of the program. This is saved in a file on program exit and
+-- | looked up from the file on program startup.
+data RunState = RunState
```
jtdaugherty: Please rename this to `LastRunState`.

jtdaugherty: (And add `Last` to function names, etc.)

src/RunState.hs
```diff
@@ -0,0 +1,78 @@
...
+
+runStateFileMode :: P.FileMode
+runStateFileMode = P.unionFileModes P.ownerReadMode P.ownerWriteMode
+
+-- | Writes the run state to a file. The file is specific to the current team.
+writeRunState :: ChatState -> IO (Either String ())
+writeRunState cs = runExceptT . convertIOException $ do
+  let runState = toRunState cs
+      tId      = cs^.csMyTeam.teamIdL
+  runStateFile <- runStateFilePath . unId . toId $ tId
```
jtdaugherty: Please don't mix "." and "$" on the same line.

src/FilePaths.hs
```diff
@@ -34,12 +38,19 @@ xdgName = "matterhorn"
 historyFileName :: FilePath
 historyFileName = "history.txt"
 
+runStateFileName :: Text -> FilePath
+runStateFileName teamId = "run_state_" ++ unpack teamId ++ ".txt"
```
jtdaugherty: Also please include a `last_` prefix here to match the renames elsewhere.

src/Main.hs
```diff
@@ -11,6 +11,7 @@ import           Config
 import           Options
 import           Types
 import           InputHistory
+import           RunState
```
jtdaugherty: Also please rename this to `LastRunState`.

src/RunState.hs
```diff
@@ -0,0 +1,78 @@
...
+  createDirectoryIfMissing True $ dropFileName runStateFile
+  writeFile runStateFile $ show runState
+  P.setFileMode runStateFile runStateFileMode
+
+-- | Reads the run state from a file given the current team ID.
+readRunState :: TeamId -> IO (Either String RunState)
+readRunState tId = runExceptT $ do
+  contents <- convertIOException $
+    (runStateFilePath . unId . toId $ tId) >>= S.readFile
+  case reads contents of
```
jtdaugherty: This a bit larger of an ask, but: can you refactor this to write a JSON object to the file rather than the `Show` instance of `RunState`? This needs to be robust to changes in `RunState`, and right now if someone adds a new field to that type, it will break all future `RunState` parses. Using a JSON object to populate the `RunState` value at parse time means we can gracefully deal with missing fields as new ones are added later and ensure that existing run state files can be loaded.

jtdaugherty:
@abhin4v - thanks very much for your work on this! The patch looks great to me overall, but I've left comments requesting some small-to-medium-sized adjustments.

abhin4v:
@jtdaugherty I have made all the changes you suggested in the comments.

jtdaugherty:
Thank you!

PR #339, opened 2017-12-08
Fix the crash on startup when the last selected channel was a direct/group message

abhin4v:
- If the last selected channel ID in the saved last_run_state file happened to be that of a direct/group message, the program crashed on startup because it couldn't find that channel in the list of channel.
- This commit fixes the bug by saving the last run state file only if the last selected channel happens to be an ordinary or private channel.
- This is how the web client behaves as well. It does not remember the last selected channel if that happens to be a direct or group message.

### DIFF
--- a/src/LastRunState.hs
+++ b/src/LastRunState.hs
@@ -13,6 +13,7 @@ module LastRunState
 import Prelude ()
 import Prelude.Compat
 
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks!

Issue #343, opened 2017-12-18
Show a typing indicator for users who are currently typing


abhin4v:
Show a typing indicator for users who are currently typing like the web client does.

jtdaugherty:
This is now merged to `develop`. Thanks again, @abhin4v!

PR #344, opened 2017-12-18
Show an indicator for users who are currently typing

abhin4v:
### Adds user typing status to the UI
- Adds per channel map to ChatState of users who are currently typing in the channel.
- Adds an UI element to show the names of users who are currently typing.
  - Shows user names if there are less than three users. Else says "several people".
- Adds an event handler for the websocket typing event to update the typing users maps.
- Adds a background thread to expire users from the typing users maps to remove the users who have stopped typing.

### Adds support for sending user_typing actions to server
- Sends the user typing actions to the server over websocket for notifying other users.
- ChatState now contains a TChan to which Websocket actions are pushed.
- when the websocket is connected in Connections.connectWebsockets, we loop forever, taking actions from the TChan in ChatState and sending them over the websocket.
- the user_typing actions are throttled so that they are sent to server only once in 2 seconds.

### Adds a config to toggle user typing indicator
- This config also toggles sending the user typing websocket actions.

Fixes Issue #343. Depends on https://github.com/matterhorn-chat/mattermost-api/pull/42.

### DIFF
--- a/sample-config.ini
+++ b/sample-config.ini
@@ -161,7 +161,8 @@
 #
 # showOlderEdits = True
 
-# Whether to show the indicator for users typing.
+# Whether to show the indicator for users typing and to send the typing
+# notifications to the server.
 # Default: False
 #
 # showTypingIndicator = False
--- a/sample-config.ini
+++ b/sample-config.ini
@@ -160,3 +160,8 @@
 # Default: True
 #
 # showOlderEdits = True
+
+# Whether to show the indicator for users typing.
+# Default: False
+#
+# showTypingIndicator = False
--- a/src/Connection.hs
+++ b/src/Connection.hs
@@ -8,21 +8,24 @@ import           Control.Concurrent (forkIO, threadDelay)
 import           Control.Concurrent.MVar
 import           Control.Exception (SomeException, catch)
 import           Control.Monad (void)
+import           Control.Monad.IO.Class (liftIO)
 import           Lens.Micro.Platform
 
 import           Network.Mattermost.WebSocket
 
 import           Types
 
-connectWebsockets :: ChatState -> IO ()
-connectWebsockets st = do
-  let shunt (Left msg) = writeBChan (st^.csResources.crEventQueue) (WebsocketParseError msg)
-      shunt (Right e) = writeBChan (st^.csResources.crEventQueue) (WSEvent e)
-      runWS = mmWithWebSocket (st^.csResources.crSession) shunt $ \ _ -> do
-                writeBChan (st^.csResources.crEventQueue) WebsocketConnect
-                waitAndQuit st
-  void $ forkIO $ runWS `catch` handleTimeout 1 st
-                        `catch` handleError 5 st
+connectWebsockets :: MH ()
+connectWebsockets = do
+  st <- use id
+  liftIO $ do
+    let shunt (Left msg) = writeBChan (st^.csResources.crEventQueue) (WebsocketParseError msg)
+        shunt (Right e) = writeBChan (st^.csResources.crEventQueue) (WSEvent e)
+        runWS = mmWithWebSocket (st^.csResources.crSession) shunt $ \ws -> do
+                  writeBChan (st^.csResources.crEventQueue) $ WebsocketConnect ws
+                  waitAndQuit st
+    void $ forkIO $ runWS `catch` handleTimeout 1 st
+                          `catch` handleError 5 st
 
 waitAndQuit :: ChatState -> IO ()
 waitAndQuit st =
--- a/src/Connection.hs
+++ b/src/Connection.hs
@@ -34,6 +34,12 @@ connectWebsockets = do
     void $ forkIO $ runWS `catch` handleTimeout 1 st
                           `catch` handleError 5 st
 
+-- | Take websocket actions from the websocket action channel in the ChatState and
+-- | send them to the server over the websocket.
+-- | Takes and propagates the action sequence number which in incremented for
+-- | each successful send.
+-- | Keeps and propagates a map of channel id to last user_typing notification send time
+-- | so that the new user_typing actions are throttled to be send only once in two seconds.
 processWebsocketActions :: ChatState -> WS.MMWebSocket -> Int64 -> HM.HashMap ChannelId (Max UTCTime) -> IO ()
 processWebsocketActions st ws s userTypingLastNotifTimeMap = do
   action <- STM.atomically $ STM.readTChan (st^.csWebsocketActionChan)
--- a/src/Connection.hs
+++ b/src/Connection.hs
@@ -10,10 +10,13 @@ import           Control.Exception (SomeException, catch)
 import           Control.Monad (void)
 import           Control.Monad.IO.Class (liftIO)
 import           Data.Int (Int64)
+import qualified Data.HashMap.Strict as HM
+import           Data.Semigroup (Max(..), (<>))
 import           Data.Time (UTCTime(..), secondsToDiffTime, getCurrentTime, diffUTCTime)
 import           Data.Time.Calendar (Day(..))
 import           Lens.Micro.Platform
 
+import           Network.Mattermost.Types (ChannelId)
 import qualified Network.Mattermost.WebSocket as WS
 
 import           Constants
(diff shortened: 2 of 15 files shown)


src/Types.hs
```diff
@@ -462,8 +463,9 @@ data Mode =
     | PostListOverlay PostListContents
     deriving (Eq)
 
--- | We're either connected or we're not.
-data ConnectionStatus = Connected | Disconnected
+-- | We're either connected or we're not. If connected, keep the latest
+-- | websocket action sequence number and the websocket itself.
+data ConnectionStatus = Connected Int64 MMWebSocket | Disconnected
```
jtdaugherty: I'd like to avoid keeping the websocket in the state and then having to update it. Instead I'd rather have a `TChan` in the state that can be used to communicate with whatever is managing the websocket. Would you be willing to refactor this so that's the case, and so that `Connected` has no arguments?

jtdaugherty:
Can you say a bit more about why the websocket sequence number now needs to be part of the state? I left a comment related to this, which is about avoiding keeping the websocket in the state at all.

sample-config.ini
```diff
@@ -160,3 +160,8 @@
 # Default: True
 #
 # showOlderEdits = True
+
+# Whether to show the indicator for users typing.
+# Default: False
```
jtdaugherty: Can you also indicate here that this also controls whether typing notifications are *sent*, too?

src/State/Setup/Threads.hs
```diff
@@ -63,6 +67,23 @@ startUserRefreshThread lock session requestChan = void $ forkIO $ forever refres
               Right upd -> return upd
           threadDelay (seconds userRefreshInterval)
 
+-- This thread refreshes the map of typing users every second, forever,
+-- to expire users who have stopped typing. Expiry time is 3 seconds.
+startTypingUsersRefreshThread :: RequestChan -> IO ()
+startTypingUsersRefreshThread requestChan = void $ forkIO $ forever refresh
+  where
+    seconds = (* (1000 * 1000))
```
jtdaugherty: This needs to be much less frequent - refreshing every second causes a screen redraw every second, so this could probably be okay at, say, 5 seconds.

src/Types.hs
abhin4v: There needs to be a new thread running to watch the `TChan` in that case, right? Is that okay? If so, I can refactor this.

abhin4v: The websocket sequence number is supposed to be a monotonically increasing number. From the mattermost documentation [here](https://api.mattermost.com/#tag/WebSocket), "The seq or sequence number is set by the client and should be incremented with every use. It is used to distinguish responses to requests that come down the WebSocket."

If this is refactored to use a separate thread to watch the `TChan` as you mentioned, we can more this sequence number to that thread instead of keeping it in the `ChatState`.

jtdaugherty: Yeah, I think that would be a nice change. The code in `Connection.hs` is already pretty close to that. I imagine that `RefreshWebsocketEvent` might go away, making `Events.hs` no longer repsonsible for calling `connectWebsockets` since that'd be done by a websocket manager thread (that doesn't exist).

jtdaugherty: > If this is refactored to use a separate thread to watch the TChan as you mentioned, we can more this sequence number to that thread instead of keeping it in the ChatState.

Yeah, that's what I had in mind. It hides that management detail from the rest of the program.

abhin4v: Sure. I'll work on this and the other comments and update the PR.

jtdaugherty: Thank you!

src/State/Setup.hs
```diff
@@ -204,6 +204,8 @@ initializeState cr myTeam myUser = do
   -- Start background worker threads:
   -- * User status refresher
   startUserRefreshThread (cr^.crUserStatusLock) session requestChan
+  -- * Refresher for users who are typing currently
+  startTypingUsersRefreshThread requestChan
```
kquick: This thread should probably not be started if the configuration parameter is disabled.

src/State/Editing.hs
```diff
@@ -190,13 +194,30 @@ handleEditingInput e = do
                         csEditState.cedEditor %= applyEdit Z.moveRight
                     | otherwise -> doInsertChar
 
-          _ | editingPermitted st -> mhHandleEventLensed (csEditState.cedEditor) handleEditorEvent e
+          _ | editingPermitted st -> do
+              mhHandleEventLensed (csEditState.cedEditor) handleEditorEvent e
+              sendUserTypingAction
```
jtdaugherty: This might be a bigger ask than some of the others, but: sending a typing action notification on every keystroke is going to result in a lot of HTTP traffic on the websocket. My concern is that the frequency of those outgoing events might be much greater than the frequency required to show the typing notification on other clients, which means a lot of outgoing traffic would be wasted. Is this how the web client implements this? What I'm wondering is whether it would be possible to abstract this a bit so that the outgoing typing notification frequency is capped at, say, N/2, where N is the expiration period of the notification on other (typical) clients. So if other clients would only need a refresh at around N seconds to keep showing the notification, we could send these out every N/2 seconds at most to keep those client notifications fresh but avoid wasting outbound traffic when someone is typing quickly. :)

jtdaugherty: On another note: this might not be the only editing handler that needs to trigger these events. Lots of things will indeed fall through to `handleEditorEvent`, but any place we've implemented other editing operations like word kill, etc., will also need to call `sendUserTypingAction`.

abhin4v: I checked the web client code and it uses 5 seconds as the expiration period (as implemented [here](https://github.com/mattermost/mattermost-webapp/blob/c41de179cdb82435f7b32e346811f7cad20603dd/stores/user_typing_store.jsx#L47) and as set [here](https://github.com/mattermost/mattermost-server/blob/7ed1177a2b676aa4c93515268642c855cfe57a37/config/default.json#L50)).

I'll change `sendUserTypingAction` to send notification to the server once every 2 seconds.

jtdaugherty: Thanks! What do you have in mind? I'm imagining a state field (per-channel) storing the last time a typing notification was sent, where `sendUserTypingAction` ignores requests unless 2 seconds have elapsed, in which case the timestamp gets updated and an action gets sent.

abhin4v: When we move to the `TChan` and a separate thread to push websocket actions, we can just add a timestamp along with the action in the channel (that is, the channel is of type `TChan (WebsocketAction, UTCTime)`) and the pusher thread can discard the actions which occur with 2 seconds of the last action's timestamp. The thread can keep this timestamp a local variable, just like it'll keep the current action sequence number. This way, we avoid adding anything into the `ChatState`.

How does this sound?

jtdaugherty: Well, with that approach I'm concerned that later there might be `WebsocketAction` values that have no need for this behavior, or for which delaying sending on a timestamp basis would be broken. The specific action of the user typing in a channel may end up being the only one that has this need.

How about this: rather than sending a raw websocket action over the channel, make a new type to abstract it (with a better name than this):
```
data CustomAction = UserTyping UTCTime ChannelId
```
and then:
* the metadata required for that specific action is required to be available when the action is sent,
* the thread can still do local state management to track when to send actions, and
* the thread can translate these higher-level application-specific notions into the lower-level websocket actions as appropriate.

jtdaugherty: (It might be better, to avoid confusion, to call the above constructor `SelfTyping` or something since it is about the user running Matterhorn rather than some other user who is typing.)

abhin4v: Alright, I'll do it this way.

jtdaugherty: Thanks!

src/State/Setup/Threads.hs
abhin4v: setting the expiry interval to 5 seconds (as the web client does) and the refresh interval to 3 seconds (half the expiry interval)

abhin4v:
@jtdaugherty i have made the changes following your suggestions.

src/Connection.hs
```diff
@@ -24,16 +27,26 @@ connectWebsockets = do
...
+    then do
+      WS.mmSendWSAction (st^.csResources.crConn) ws $ convert action
+      processWebsocketActions st ws (s + 1) =<< getCurrentTime
+    else do
+      processWebsocketActions st ws s lastUserTypingNotifTime
   where
-    convert (UserTyping cId pId) = WS.UserTyping s cId pId
+    convert (UserTyping _ cId pId) = WS.UserTyping s cId pId
+
+    shouldSendAction (UserTyping ts _ _) =
```
jtdaugherty: This doesn't appear to track the channel ID associated with the last typing notification time, so it means that if I type in channel A and switch to channel B in less than two seconds and start typing, the typing in channel B will not cause a server notification of typing to occur.

src/Types.hs
```diff
@@ -518,7 +520,8 @@ newState rs i u m tz hist sp = ChatState
   , _csChannelSelectState          = emptyChannelSelectState
   , _csRecentChannel               = Nothing
   , _csUrlList                     = list UrlList mempty 2
-  , _csConnectionStatus            = Disconnected
+  , _csConnectionStatus            = Connected
```
jtdaugherty: Why did the initial connection status change to `Connected`? I think we rely on the `Disconnected` -> `Connected` transition at startup.

jtdaugherty:
Thanks! I've left a few more comments.

src/Connection.hs
abhin4v: I'll fix this.

src/Types.hs
abhin4v: It is set to `Connected` in the develop branch https://github.com/matterhorn-chat/matterhorn/blob/develop/src/Types.hs#L518. I had changed it `Disconnected` in a previous commit but I changed it back in this commit.

jtdaugherty: Oh, I see - thanks!

src/Connection.hs
jtdaugherty: Thanks!

jtdaugherty:
One other thing, @abhin4v - with my latest feedback addressed, this is about ready to merge. Would you be willing to do a pass over the things you have added and make sure that all new functions and types have Haddock comments?

abhin4v:
@jtdaugherty Fixed the issues and added documentation.

jtdaugherty:
Thanks for all of your work on this, and thanks for all the updates in response to our feedback!

abhin4v:
@jtdaugherty I was wondering if the `WebsocketActionChan` should be moved into `ChatResources` from `ChatState`. What do you think?

jtdaugherty:
Yeah, that sounds good to me - the intent with `ChatResources` was for things that don't change for the lifetime of the program.

PR #345, opened 2017-12-18
Add caching to travis config for faster builds

abhin4v:
Adds caching for cabal and ghc directories to reduce build time on travis. Reduces each job's build time from around 35 minutes to around 5 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,6 @@ script:
 
 cache:
   directories:
-  - dist-newstyle
   - $HOME/.cabal
   - $HOME/.ghc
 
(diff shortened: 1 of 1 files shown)


jtdaugherty:
I'd like to avoid caching `dist-newstyle` because caching it can lead to build successes that aren't representative (i.e. a cached `dist-newstyle` can lead to a successful build that should otherwise fail due to changes in the `cabal` file or Hackage).

abhin4v:
@jtdaugherty I have removed `dist-newstyle` from the list.

jtdaugherty:
Thank you!

PR #347, opened 2017-12-20
Make get calls to server concurrent for faster startup

abhin4v:
- does this only for independent calls.
- inlines `mmGetAllChannelsWithDataForUser` into `refreshChannelsAndUsers` and makes all the get calls concurrent.

### DIFF
--- a/matterhorn.cabal
+++ b/matterhorn.cabal
@@ -119,7 +119,8 @@ executable matterhorn
                      , skylighting          >= 0.3.3.1 && < 0.4
                      , timezone-olson       >= 0.1.7   && < 0.2
                      , timezone-series      >= 0.1.6.1 && < 0.2
-                     , aeson                >= 1.2.3.0 && < 1.3
+                     , aeson                >= 1.1     && < 1.3
+                     , async                >= 2.0     && < 2.2
   default-language:    Haskell2010
 
 test-suite test_messages
--- a/matterhorn.cabal
+++ b/matterhorn.cabal
@@ -119,7 +119,7 @@ executable matterhorn
                      , skylighting          >= 0.3.3.1 && < 0.4
                      , timezone-olson       >= 0.1.7   && < 0.2
                      , timezone-series      >= 0.1.6.1 && < 0.2
-                     , aeson                >= 1.1     && < 1.3
+                     , aeson                >= 1.2.3.0 && < 1.3
                      , async                >= 2.0     && < 2.2
   default-language:    Haskell2010
 
--- a/src/State.hs
+++ b/src/State.hs
@@ -281,6 +281,8 @@ updatePreference pref = do
 -- occurs.
 refreshChannelsAndUsers :: MH ()
 refreshChannelsAndUsers = do
+  -- The below code is a duplicate of mmGetAllChannelsWithDataForUser function,
+  -- which has been inlined here to gain a concurrency benefit.
   session <- use (csResources.crSession)
   myTeamId <- use (csMyTeam.teamIdL)
   myId <- use (csMe.userIdL)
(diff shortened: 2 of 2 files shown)


matterhorn.cabal
```diff
@@ -119,7 +119,8 @@ executable matterhorn
                      , skylighting          >= 0.3.3.1 && < 0.4
                      , timezone-olson       >= 0.1.7   && < 0.2
                      , timezone-series      >= 0.1.6.1 && < 0.2
-                     , aeson                >= 1.2.3.0 && < 1.3
```
jtdaugherty: Also, was there a reason for changing the Aeson bounds?

jtdaugherty: On a related note, we haven't documented this really explicitly, but we tend to use tight bounds on all dependencies to ensure build stability. We aren't optimizing for allowing the average Hackage user to build Matterhorn against a wide variety of dependency versions. If there are ever changes to the dependency bounds, they need to be well-motivated and well-tested: if we change bounds because we need to, the commit should provide the justification, and if we increase upper bounds to permit new versions, we should also increase lower bounds to ensure that the version space continues to be constrained.

jtdaugherty: This pull request doesn't seem to change anything related to Aeson functionality, so if you wanted to change the Aeson bounds, I'd like to see that in a separate commit or pull request.

abhin4v: This was by mistake. I reverted it.

abhin4v:
@jtdaugherty added a comment about duplication.

matterhorn.cabal
jtdaugherty: Okay, thank you!

src/State.hs
```diff
@@ -280,12 +281,21 @@ updatePreference pref = do
...
 refreshChannelsAndUsers = do
+  -- The below code is a duplicate of mmGetAllChannelsWithDataForUser function,
+  -- which has been inlined here to gain a concurrency benefit.
   session <- use (csResources.crSession)
   myTeamId <- use (csMyTeam.teamIdL)
   myId <- use (csMe.userIdL)
   doAsyncWith Preempt $ do
-    chansWithData <- mmGetAllChannelsWithDataForUser session myTeamId myId
-    uMap <- mmGetProfiles session myTeamId 0 10000
+    (chans, datas, uMap) <- runConcurrently $ (,,)
```
kquick: As a minor note, adding concurrency "under" the work queue like this (which is well done here) will cause incorrect reporting of background async work counts (https://github.com/matterhorn-chat/matterhorn/blob/master/src/Draw/Main.hs#L585).  That's not really an issue for this change, but something to keep in mind if more of this style is used.  

The background work counts are more of a debugging aid anyhow, but the utility will be less if they become consistently inaccurate.  I don't think it's worth adding *significant* complexity to maintain the accuracy; just something to make a deliberate decision on if more of this concurrency style is introduced in the future.

That said, I like this change, thank you @abhin4v !

jtdaugherty: Thanks, @kquick - I thought the work count was the queue length, and this doesn't change that, right? How does it become inaccurate?

kquick: Right, the work count is the queue length, but as I understand this PR, this single work queue item will end up issuing 3 requests to the server.  Previously there was a 1:1 correlation between queue entry and server request, so I suppose it depends on what the "work count" means, but it's probably most useful as a general indicator of pending server interactions (i.e. the "slow" stuff).

jtdaugherty: > Previously there was a 1:1 correlation between queue entry and server request

That isn't the case; there are quite a few async work queue items that make multiple (serial) requests to the server in the course of their work.

kquick: Ahh, right.  And I didn't have to look very far to find one and remember the pattern: https://github.com/matterhorn-chat/matterhorn/blob/master/src/State.hs#L227-L232.  I updated the description to keep myself from getting confused again.  :-)  https://github.com/matterhorn-chat/matterhorn/commit/7fce525d9c6c4a8a50f491a8e2088bfa59c4311f

PR #357, opened 2018-02-22
Optionally use nicknames as display names

kellymclaughlin:
Based on the mattermost server settings use users' nicknames as display names
and fall back to using the mattermost username if a user has not set a
nickname. When enabled on the server, nicknames may also be used in tab completions. The behavior mimics that of the mattermost desktop client where the name filled in for the tab completion is the actual username. It feels a little odd, but that's how the official client does it so that's what I went with.

The limitation with this is that changes to the server setting for display name do not generate a websocket event and therefore matterhorn will not detect a change to the setting until it is restarted. The mattermost desktop client appears to get around this by sending RSTs on the mattermost sockets and re-initializing. In practice I suspect this setting is not changed very frequently and will not cause a problem for many matterhorn users.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -71,6 +71,7 @@ module Types
   , csChannels
   , csChannelSelectState
   , csEditState
+  , csClientConfig
   , timeZone
   , whenMode
   , setMode
@@ -169,6 +170,8 @@ module Types
   , addNewUser
   , setUserIdSet
   , channelMentionCount
+  , useNickname
+  , displaynameForUserId
 
   , userSigil
   , normalChannelSigil
@@ -189,6 +192,7 @@ import           Brick.BChan
 import           Brick.AttrMap (AttrMap)
 import           Brick.Widgets.Edit (Editor, editor)
 import           Brick.Widgets.List (List, list)
+import           Control.Applicative ((<|>))
 import           Control.Monad (when)
 import qualified Control.Concurrent.STM as STM
 import           Control.Concurrent.MVar (MVar)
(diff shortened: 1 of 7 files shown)


src/Types.hs
```diff
@@ -71,6 +71,8 @@ module Types
   , csChannels
   , csChannelSelectState
   , csEditState
+  , csClientConfig
+  , csUsers
```
jtdaugherty: We underwent a recent refactor to avoid exporting `csUsers` and many other lenses. There are now various functions to get the underlying data in a way that conceals the underlying representation. Can you use those? In `Types`, see e.g. `allUserIds`, `usernameForUserId`, `userIdForUsername`, etc. - there's a long list of such accessors in the export list now. (Those are very new relative to your last pass at doing this work.)

jtdaugherty:
Thank you! Overall it looks good to me, but I left one comment to be addressed. Also, I'll have to chat with the team and figure out when it makes sense to merge this, as `develop` has some other work in testing right now.

kellymclaughlin:
Sure, I'll have a look at those new functions and incorporate them. I also realized when I ported to the work over to the latest tip of develop that channel switching doesn't work. I'll fix that up too and get it added.

jtdaugherty:
Okay. Thanks! I know `develop` has changed a lot since your last pass at this, so we appreciate your willingness to track it.

jtdaugherty:
By the way, since I mentioned that we can't merge this quite yet: once you've made changes to address my comment, the Matterhorn development team will take care of any further changes required to merge this. Since we can't merge it right now, it's possible `develop` will drift before we can, and we'll take it from there. Thanks again for the patch!

kellymclaughlin:
I pushed the fix for the channel switching as well as the changes you suggested. Thanks for the offer to take on the maintenance while other work is going on! Feel free to let me know if there's anything I can do to help with it.

src/Types/Users.hs
```diff
@@ -31,6 +32,7 @@ module Types.Users
   )
 where
 
+import           Data.Foldable            (find)
```
jtdaugherty: A tiny, tiny nit: but the whitespace convention on this import doesn't match the adjacent ones. I'll fix it if you don't get to it. (Granted, we don't do this consistently across the codebase, largely because different team members have different preferences / muscle memory.)

jtdaugherty:
Thanks for the updates!

src/Types/Users.hs
kellymclaughlin: Oh sorry I missed that. Pushed a fix for that.

jtdaugherty:
Thank you!

PR #379, opened 2018-03-23
GHC 8.4.1 support

TomMD:
This doesn't fix everything matterhorn related, just the modules.  There are dependencies that don't work out.  To test the changes I built with `cabal new-build --allow-newer`.

Would we want to maintain support for GHC < 8.4?  I dislike `-XCPP` myself.

### DIFF
--- a/src/State.hs
+++ b/src/State.hs
@@ -131,7 +131,6 @@ import qualified Data.HashMap.Strict as HM
 import qualified Data.Sequence as Seq
 import           Data.List (sort, findIndex)
 import           Data.Maybe (isJust, fromJust, catMaybes, isNothing)
-import           Data.Monoid ((<>))
 import qualified Data.Set as Set
 import qualified Data.Text as T
 import           Data.Time (getCurrentTime)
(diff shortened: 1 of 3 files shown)


jtdaugherty:
I'm not sure we can merge this yet, but thanks for doing it! We'll have to discuss whether we want to support 8.4 and when, and whether we care about supporting older GHCs at that point.

aisamanra:
I took a stab at it and was able to build to get Matterhorn to build on 8.4 without sacrificing backwards compatibility. This involved two things:
1. we already use [base-compat](http://hackage.haskell.org/package/base-compat), which does the CPP for us! That means that all instances of `import Data.Monoid` that just bring `<>` into scope can be turned into `import Data.Monoid.Compat`, which is specifically designed so that it'll export `<>` pre-8.4 and won't after it.
2. taking our two `Monoid` instances and turning them into standalone named functions. I'd argue that neither of the `Monoid`s we define actually pull their weight: the `Monoid` instance for `DirectionalSeq` was used on three lines (once as `<>` and twice as `mempty`), and the `Monoid` instance for `PostProcessMessageAdd` was used on two lines (once as `mappend` and once as a fold.) All of these could be replaced a bit more verbosely but a bit more future-proof-ly with named functions (e.g. `emptyDirectionalSeq` instead of `mempty`) and it preserves our ability to build on past versions.

After doing this locally, I tried building with every compiler version I have locally, and I was able to build on `8.{0,2,4}`, but I can't build in `7.10` because of a problem with `brick-0.35`. I can contribute a fix for that problem, but I also think it's worth asking first whether we want to still support `7.10` or if we're okay bumping the minimum version up to `8.0`.

jtdaugherty:
Actually I fixed `brick-0.35` and the 0.34.* versions to build on 7.10 yesterday.  On the minimum version: I'm fine to just bump the minimum version to 8.0 anyway.

When I tried an 8.4 build a while back, there was a Hanquez lib that didn't build due to a semigroup instance issue. Presumably you didn't run into that? If so, we should figure out how to pin the deps so everyone can enjoy that build success..

Thank you, Getty!

jtdaugherty:
> we should figure out how to pin the deps

By which I just mean, figure out which version that worked for you and depend on that.

aisamanra:
I didn't have the Hanquez dep problem this time, so presumably that one's been fixed.

When you say you fixed `brick` to build on 7.10, did you also push them to Hackage? 'Cause I'm still having problems due to `brick` not depending on `semigroup` for pre-8.0 builds.

jtdaugherty:
I updated the metadata in-place on Hackage for 0.34+ and Travis shows at least `master` building on 7.10.3.

jtdaugherty:
Interesting - 0.35 doesn't have the semigroup dependency I thought I added. I'm fixing.

jtdaugherty:
Well, it turns out that *none* of the conditional semigroup deps I added even got saved.

jtdaugherty:
Oh, it turns out there were errors that were not obvious at all, and that Hackage does not permit `if` conditionals in revisions. ;(

jtdaugherty:
A new version, `0.35.1`, is now uploaded that will build on GHC >= 7.10.3. (Probably other 7.10s, too, but I haven't tested.)

PR #387, opened 2018-04-18
Allow notifications via an external command

jasom:
This adds the configuration field `activityNotifyCmd` which can be set to a shell command that takes the following 3 positional arguments:

1. "1"/"0" if the user was/was not mentioned
2. The username of the sender
3. The text of the message

The shell command is only called for new messages, not for messages read from the backlog.

### DIFF
--- a/notify
+++ b/notify
@@ -14,7 +14,8 @@ message="${3?}"
 # First word is urgency for items where you are not mentioned, the second is
 # urgency for items where you are mentioned.  Use "none" to not be notified,
 # otherwise low, normal, critical are options
-notify_URGENCIES="none normal"
+notify_URGENCIES="normal normal"
+#notify_URGENCIES="none normal"
 # The desktop notification category
 notify_CATEGORY="im.received"
 # Notification header
@@ -24,7 +25,7 @@ notify_BODY="$message"
 
 getUrgencyHelper() {
     shift "$mentioned"
-    printf '%s' "$1"
+    echo "$1"
 }
 
 getUrgency() {
--- a/notify
+++ b/notify
@@ -1,0 +1,37 @@
+#!/bin/sh
+#############################################################
+# Sample shell-script for using notify-send with matterhorn #
+#############################################################
+
+# Positional parameters
+mentioned="${1?}"
+sender="${2?}"
+message="${3?}"
+
+# Script options
+
+# notify_URGENCIES
+# First word is urgency for items where you are not mentioned, the second is
+# urgency for items where you are mentioned.  Use "none" to not be notified,
+# otherwise low, normal, critical are options
+notify_URGENCIES="none normal"
+# The desktop notification category
+notify_CATEGORY="im.received"
+# Notification header
+notify_HEAD="Matterhorn message from $sender"
+# Notification body
+notify_BODY="$message"
+
+getUrgencyHelper() {
+    shift "$mentioned"
+    printf '%s' "$1"
+}
+
+getUrgency() {
+    # we are using arguments as a poor-mans bash array for portability
+    # shellcheck disable=SC2086
+    getUrgencyHelper $notify_URGENCIES
+}
+
+test "$(getUrgency)" = "none" ||
+    notify-send -u "$(getUrgency)" -c "$notify_CATEGORY" "$notify_HEAD" "$notify_BODY"
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -55,7 +55,7 @@ fromIni = do
       (configShowTypingIndicator defaultConfig)
     configEnableAspell <- fieldFlagDef "enableAspell"
       (configEnableAspell defaultConfig)
-    configActivityNotifyCmd <- fieldMb "activityNotifyCmd"
+    configActivityNotifyCommand <- fieldMb "activityNotifyCommand"
     configActivityBell <- fieldFlagDef "activityBell"
       (configActivityBell defaultConfig)
     configHyperlinkingMode <- fieldFlagDef "hyperlinkURLs"
@@ -121,7 +121,7 @@ defaultConfig =
            , configSmartBacktick             = True
            , configURLOpenCommand            = Nothing
            , configURLOpenCommandInteractive = False
-           , configActivityNotifyCmd         = Nothing
+           , configActivityNotifyCommand     = Nothing
            , configActivityBell              = False
            , configShowBackground            = Disabled
            , configShowMessagePreview        = False
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -55,6 +55,7 @@ fromIni = do
       (configShowTypingIndicator defaultConfig)
     configEnableAspell <- fieldFlagDef "enableAspell"
       (configEnableAspell defaultConfig)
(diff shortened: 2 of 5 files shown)


jasom:
This is my first haskell I've written since I worked through a tutorial about a decade ago, so please read closely and feel free to correct any style issues.

jasom:
I will rebase off of develop.

I'll have to test multi-line messages, but assuming they are just a single Text object they should be passed unmolested.

jasom:
@jtdaugherty I've rebased off of develop

jtdaugherty:
Thank you!

src/Config.hs
```diff
@@ -55,6 +55,7 @@ fromIni = do
       (configShowTypingIndicator defaultConfig)
     configEnableAspell <- fieldFlagDef "enableAspell"
       (configEnableAspell defaultConfig)
+    configActivityNotifyCmd <- fieldMb "activityNotifyCmd"
```
jtdaugherty: Please rename the setting name to `activityNotifyCommand` (just the configuration setting string, not the record field name).

Also, can you please add a section on this new setting to `sample-config.ini`?

src/State.hs
```diff
@@ -1286,6 +1286,31 @@ deleteMessage new = do
   cId <- use csCurrentChannelId
   when (postChannelId new == cId) updateViewed
 
+maybePostUsername :: ChatState -> Post -> T.Text
+maybePostUsername st p = case do { uId <- postUserId p; name <- usernameForUserId uId st; return name } of
+    Nothing -> T.empty
+    Just x -> x
```
jtdaugherty: Some style notes:
* The binding of (and then `return` of) `name` is redundant.
* This whole thing can be shortened to:
```
fromMaybe T.empty $ do
  uId <- postUserId p
  usernameForUserId uId st
```

src/State.hs
```diff
@@ -1286,6 +1286,31 @@ deleteMessage new = do
...
+
+maybeNotify :: [PostToAdd] -> MH ()
```
jtdaugherty: The `maybeNotify` function is written to work on lists of `PostToAdd`, but it should be written more simply to just run on a single `PostToAdd` and then the call site in `postOp` would instead say `mapM_ maybeNotify p`.

src/State.hs
```diff
@@ -1286,6 +1286,31 @@ deleteMessage new = do
...
+maybeNotify [] = do 
+    return ()
+maybeNotify (OldPost _:ps) =
+    maybeNotify ps
+maybeNotify (RecentPost p m:ps) = do
```
jtdaugherty: Can you factor the body of this case out as its own function? That will help clarify what is being done here.

jtdaugherty: Also, can you use more descriptive names than `m` and `p` here?

src/State.hs
```diff
@@ -1286,6 +1286,31 @@ deleteMessage new = do
...
+    outputChan <- use (csResources.crSubprocessLog)
+    st <- use id
+    notifyCmd <- use (csResources.crConfiguration.to configActivityNotifyCmd)
+    case notifyCmd of
+        Nothing -> return ()
+        Just cmd ->
+            doAsyncWith Preempt $ do
+                let (messageString, notified, sender) = (T.unpack (postMessage p), if m then "1" else "2",  T.unpack (maybePostUsername st p))
```
jtdaugherty: This line is a little too long. Generally we'd write this as three separate `let` bindings instead of one using a triple, i.e.,
```
let messageString = T.unpack $ postMessage p
    notified = if m then "1" else "2"
    sender = T.unpack $ maybePostUsername st p
```

src/State.hs
```diff
@@ -1299,28 +1324,30 @@ maybeRingBell = do
...
-andProcessWith x NoAction                    = x
-andProcessWith _ x                           = x
+andProcessWith NoAction x                                        = x
+andProcessWith (NotifyUserAndServer p) UpdateServerViewed        = NotifyUserAndServer p
+andProcessWith (NotifyUserAndServer p1) (NotifyUser p2)          = NotifyUserAndServer (p1 <> p2)
+andProcessWith (NotifyUserAndServer p1) (NotifyUserAndServer p2) = NotifyUserAndServer (p1 <> p2)
+andProcessWith (NotifyUser p1) (NotifyUser p2)                   = NotifyUser (p1 <> p2)
+andProcessWith (NotifyUser p) UpdateServerViewed                 = NotifyUserAndServer p
+andProcessWith UpdateServerViewed UpdateServerViewed             = UpdateServerViewed
+andProcessWith x y                                               = andProcessWith y x
```
jtdaugherty: This last recursive case that just swaps the order is dangerous, because if someone later adds constructors to the `PostProcessMessageAdd` type but forgets to update the various cases above this one, this can cause an infinite loop. Instead I'd rather have the possible cases enumerated so we can remove this case entirely, because this will cause GHC to warn us when we fail to enumerate the possible cases.

src/Types.hs
```diff
@@ -268,6 +268,7 @@ data Config = Config
   , configSmartBacktick             :: Bool
   , configURLOpenCommand            :: Maybe Text
   , configURLOpenCommandInteractive :: Bool
+  , configActivityNotifyCmd         :: Maybe T.Text
```
jtdaugherty: On second thought, please expand `Cmd` to `Command` here, too, just as with the setting name, to be consistent with the other settings that hold commands.

jtdaugherty:
Okay, I'm done leaving comments on the current version.

notify
```diff
@@ -0,0 +1,37 @@
...
+# The desktop notification category
+notify_CATEGORY="im.received"
+# Notification header
+notify_HEAD="Matterhorn message from $sender"
+# Notification body
+notify_BODY="$message"
+
+getUrgencyHelper() {
+    shift "$mentioned"
+    printf '%s' "$1"
```
kquick: I don't think that the original Bourne shell had `printf` support, so it's likely not available in other derivatives as well.  Your comment below indicates you are attempting to be portable, but since this is a captive script and not the user's main environment, I'd suggest it's probably fine to just explicitly use `#!/bin/bash`.  Alternatively, I think just `echo "$1"` would be safe here.

jtdaugherty: If we're going to go with `bash`, this is more portable: `#!/usr/bin/env bash`.

jasom: printf is quite portable, even without shell support (see e.g. `/usr/bin/printf`) but you are correct that echo is safe here; I have a habit of not using `echo "$foo"` since that breaks in various ways when `$foo` begins with a `-`

jasom:
I believe I made all suggested changes.  I won't likely have a chance to look at any further comments for ~14 hours.

jtdaugherty:
Thanks! We'll take a look.

kquick:
Thanks @jasom! There's still the concern about infinite recursion on `andProcessWith` though.

jtdaugherty:
Aside from @kquick's mention of the outstanding recursion concern, I'm happy with the other changes. Thanks!

jasom:
good catch @kquick, this should be fixed now

jtdaugherty:
Thank you for the patch, and thanks for making adjustments from our feedback!

PR #406, opened 2018-08-03
A simple random video chat URL generator.

TomMD:
I used openssl believing that was most wide spread. No reason we couldn't instead or in addition detect and use base64 or any other solution.

### DIFF
--- a/scripts/talky
+++ b/scripts/talky
@@ -2,6 +2,17 @@
 
 BASE_URL="https://talky.io"
 
+base64enc() {
+    if [[ -x $(which openssl) ]] ; then
+        retval=$(echo $1 | openssl enc -base64)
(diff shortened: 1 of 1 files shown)


scripts/talky
```diff
@@ -2,6 +2,17 @@
 
 BASE_URL="https://talky.io"
 
+base64enc() {
+    if [[ -x $(which openssl) ]] ; then
+        retval=$(echo $1 | openssl enc -base64)
+    if [[ -x $(which base64) ]] ; then
+        retval=$(echo $1 | base64)
+    else
+        echo "Unable to base64 encode, install base64?"
```
jtdaugherty: It should probably `exit` here instead, since otherwise this error message will make it through to `matterhorn`.

jtdaugherty:
Thanks!

PR #410, opened 2018-08-22
Talky.io doesn't like '/' and perhaps other base64 symbols

TomMD:
It isn't surprising '/' is a problem. It is surprising it is silently a problem and both users end up in a chat that is (evidently) non-joinable.

### DIFF
--- a/scripts/talky
+++ b/scripts/talky
@@ -3,7 +3,9 @@
 BASE_URL="https://talky.io"
 
 base64enc() {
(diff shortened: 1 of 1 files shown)


jtdaugherty:
@TomMD can you please update this to use a SHA variant? If you don't have time, let me know. Thanks!

TomMD:
@jtdaugherty Sure, done as yet another option for random but web route friendly string generation.

jtdaugherty:
Thanks!

PR #413, opened 2018-09-13
Use git rev-parse in install.sh

grandchild:
Removes sed line which is hard to comprehend and might fail if git decides to change "branch" output formatting.


jtdaugherty:
Thanks!

PR #481, opened 2019-01-16
fix osx compilation

MartinDelille:
I manage compile matterhorn on osx by amending the `build.sh` script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@
 # https://github.com/hvr/multi-ghc-travis
 language: c
 sudo: false
+os:
+  - osx
+  - linux
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


jtdaugherty:
Thanks for the patch! Right now we need to standardize on a version of GHC and `cabal-install` on OS X, and this patch skips ahead of that a bit and also assumes the presence of Homebrew. I'm glad you got your build working, but I'd like to hold off on this kind of change until we can formalize our toolchain requirements.

MartinDelille:
Ok no worries. I wrote an [Homebrew](https://brew.sh) formula to make the installation easy on *OSX*: https://github.com/MartinDelille/homebrew-tap/blob/master/matterhorn.rb

*OXS* user can install installing this way:

```
brew tap martindelille/tap
brew install matterhorn
```

Do you want me to push this doc somewhere?

PR #513, opened 2019-08-12
Add attachments path to config

wiruzx:


jtdaugherty:
Thank you for the patch! I have a couple of requests:
* Please rename the setting to `defaultAttachmentPath`.
* Please update `sample-config.ini` with a section for the new setting, commented out like the rest, describing the default behavior (if it is omitted) as well as the behavior if the value is specified.

jtdaugherty:
Also, can you please test the behavior of the application if the configuration-specified attachment path is invalid or unreadable?

wiruzx:
@jtdaugherty Sure!

wiruzx:
@jtdaugherty Hey,
I've changed the name and added a note to `sample-config.ini`
As for testing the behavior (I assume you meant manually testing, not the unit tests), if path doesn't exist, you'll see the attachments browser with `..`, but when you select it, browser will close, as if you hit Escape key.
I guess I could add the check if the folder exists before opening attachments browser, what do you think?

jtdaugherty:
Thanks for the extra changes!

As for the behavior when the specified directory cannot be used, I think a graceful fallback is to use the current directory (i.e. the behavior when the config setting is not specified). That fallback behavior should also be documented.

wiruzx:
@jtdaugherty Could you please check last commit?
I've added the check if directory exists and added a note to the sample-config.ini

jtdaugherty:
@wiruzx thanks! The changes mention that the fallback path is the matterhorn binary path. But that's not right: the fallback path is the CWD of the `matterhorn` process.

wiruzx:
Fair enough. Will update the doc

jtdaugherty:
Thank you!

sample-config.ini
```diff
@@ -237,9 +237,8 @@
 
 # The default attachment path. Optional.
 #
-# If specified and directory with this path exists, will be used as
-# the current folder when you attach files. Otherwise the location of
-# matterhorn binary will be used
-#
+# If specified and directory at this path exists, will be used as the
+# current folder when you attach files. Otherwise the current working
+# directory of matterhorn process will be used
```
jtdaugherty: Thanks!

jtdaugherty:
Thanks, @wiruzx!

PR #547, opened 2020-01-06
[WIP] Implement team switching and multi-team support

aisamanra:
I'm creating this as a WIP PR at present because it's not ready to be merged in its current state, but I wanted to have some conversations about both the desired user-facing UI for this as well as the changes to the internal data structures necessary. I've implemented this as naïvely as possible just to prove that it works, so there are some glaring inefficiencies and some remaining bugs. (There are also a few ways to split this up into smaller PRs which might be worth doing: for example, getting rid of the threaded `TeamId` parameter in the async machinery was trivial, since it wasn't actually used anywhere, so I'd be happy to split things like that into smaller pre-work PRs to make them easier to review!)

The high-level goal of this PR is to build in support for switching between teams within a single Matterhorn session. This is how the web client works with multiple teams, as well: you can have a single session in which you switch back and forth between teams by clicking on teams on a vertical bar to the left. Each public or private channel belongs to exactly one team, while DM and group channels don't belong to any team.  When you switch to another team, you'll see a different set of public/private channels to the left, but the same set of DM channels.

Open questions I have are:
- How should this PR be broken up or staged? Should some of the pre-work (like removing the `TeamId` async machinery and adding `TeamId` metadata to the `ChannelInfo` struct) go in first, separate from these changes, to make the PRs cleaner and easier to review?
- What is the desired UI for this? For this proof-of-concept, I've included a single keybinding (currently defaulting to `C-t`) that cycles to the next one, and it'd be trivial to add a corresponding cycle-to-previous binding as well. It only shows the currently loaded team at the top, but I can imagine that if you are part of multiple teams you might want to have a bar at the top listing all the teams as well as which you're currently viewing, like
    ```
      team-one   [team-two]   team-three
    ```
    (with that horizontal line being omitted if you only belong to one team). I can also imagine wanting to choose from a list of other teams, maybe with a UI like channel selection.
- Is it worth visually indicating that activity has happened on another team? (I don't actually know how the web client does this, but it's probably worth investigating.)
- Does this change imply that team selection dialogue/the team configuration option is deprecated, or is it used for something else, like opening a default team? Should there be a way of storing the 'last open' team, as well? (For this proof-of-concept, I just straight-up deleted the team selection dialogue—why choose one when you'll have quick access to all of them?—but it might be worth keeping to choose a default team or to modify that view to use as a team-switcher instead.)
- What's the ideal data structure to be used here? For ease of implementation, this proof-of-concept simply wrapped a `HashMap TeamId ...` around the outside of the channel focus zipper, stores the channel _data_ all in the same structure (since channels have UUIDs and won't conflict across teams) and keeps around a lens that grabs the channel focus zipper for whatever the current team is, which let a bunch of code remain unchanged.
- What should the pre-fetch behavior be like? Do we only pre-fetch the channel for the team we enter at startup, or should we pre-fetch one channel for each team? (I've sidestepped this question in this specific proof-of-concept by ignoring pre-fetching and fetching everything, but that's something that needs to be changed before this is even remotely ready for merging, because it's slow and inefficient.)

With answers or preferences around the questions, I'll be happy to implement them and get this PR into a state where it could be considered for inclusion. Thank you!


jtdaugherty:
Awesome, thanks for starting this, @aisamanra! Do you think you would be able to have any in-person discussions with us about this in the next few weeks? (Before 5 PM preferable, but after 5 PM may work if the others can also be involved.)

jtdaugherty:
(And as far as an in-person meeting is concerned, my schedule is a bit full for the next two weeks, but if the timing is right, I have some availability. I'm not at all sure about @kquick or @immindich.)

kquick:
In person would be great for me, my schedule is of moderate fullness.

Another interesting item to think about: the direct user channels are somewhat independent of the team.  I've noticed this by having a different Matterhorn instance open for each team.  The visibility of users is controlled by users in the team, but if UserA is on both teams, then any messages in your direct user channel to UserA will highlight, regardless of which team you are viewing (i.e. both my Matterhorn instances highlighted that UserA channel as having unread messages).  Likewise, viewing  the UserA channel caused the highlighting to go away from both instances.

I think the primary implications of this are for notifications for unread messages relative to teams.  Alternatively, instead of having a "team" tab, we could utilize the newer channel list features @jtdaugherty  implemented recently to group public channels by team; I actually don't necessarily like this direction, but it's worth an explicit decision against this rather than overlooking it.

kquick:
Also: yes, I'd much prefer smaller steps like breaking out the TeamId changes to a separate PR... if this is the direction we go.

aisamanra:
Absolutely! I'm not sure when the best time would be, but I've been meaning to make time for a Galois lunch visit soon, at least, so maybe I can swing it in to there. I'll send a ping before I do. I've already taken some of the trivial-and-downsideless changes and put them into #548, as well.

MartinDelille:
I would definitely love to see this PR finalized!

Team switching has to be done with the mouse in the electron desktop client which is not very convenient.

I'll be happy to test it if it can help!

jtdaugherty:
@aisamanra something @kquick and I discussed today was the question of what it might take to extend this to supporting multiple server connections while we're at it. Unlike the web client, whose logic is served up by the server and probably needs to match the server's version more closely, matterhorn is not bound by that restriction and could definitely successfully interact with different server versions simultaneously (with no special client-side logic to do so). So as we think about multi-team support, I'd love to keep this idea in mind to see what it would take to pull off, and to see what the design impact would be.

jtdaugherty:
@aisamanra What's your availability like over the next few weeks? Would it be possible for us to find a time to get together during business hours? Or, failing that, around 5 PM Pacific?

jtdaugherty:
I'm going to close this for a couple of reasons. But *please* don't interpret this as a lack of desire to ultimately have this feature! I'd like to close since there has been a lot of activity since this was created and it will not likely be merged, also because it was not framed as a patch to be merged but as a place for discussion (and we have other mechanisms for that), and finally because I don't think we'll be able to dedicate attention to this soon. Thank you for your work on this, @aisamanra!

aisamanra:
I too haven't been able to work on it to the degree that I'd hoped, so that makes perfect sense. At some point I'll resurrect this PR and see what I can do with it, but I can't promise it'll be any time soon!

jtdaugherty:
As of this week, the `experimental/multi-team-support` branch has a pretty feature-complete implementation of support for multiple teams. @aisamanra if you check it out, please let us know if you find any issues!

PR #548, opened 2020-01-06
Remove TeamId parameter from async API

aisamanra:
This was some prework for #547 but also is minor cleanup that shouldn't hurt anything. It turns out that the `TeamId` parameter in the async machinery is never used—I assume this is because the `TeamId` was IIRC more widely used in v3 of the Mattermost API but became less common in v4, so this API never quite got simplified accordingly.

### DIFF
--- a/src/State/Async.hs
+++ b/src/State/Async.hs
@@ -123,7 +123,7 @@ doAsyncWithIO prio st act = do
 -- main (brick) thread.
 doAsyncMM :: AsyncPriority
           -- ^ the priority for this async operation
-          -> (Session -> TeamId -> IO a)
+          -> (Session -> IO a)
           -- ^ the async MM channel-based IO operation
           -> (a -> Maybe (MH ()))
           -- ^ function to process the results in brick event handling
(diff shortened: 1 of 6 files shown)


jtdaugherty:
This looks good to me, but I just have a question first. I haven't looked at your other work on supporting multiple teams, but why is it possible for this to go away in the async API? Is it just because any operation that needs to specify a team ID does so in an `IO` computation unbeknownst to the async machinery? (My gut just says: I thought this was needed and am surprised it can go away without any impact.)

aisamanra:
I think it was originally included in the async functions because v3 of the API used the team ID a _lot_ more pervasively than v4 of the API, especially as a URL parameter. For example, compare the [v3 "view channel" endpoint](https://api.mattermost.com/v3.7/#tag/channels%2Fpaths%2F~1teams~1%7Bteam_id%7D~1channels~1view%2Fpost) with the [v4 "view channel" endpoint](https://api.mattermost.com/#tag/channels%2Fpaths%2F~1channels~1members~1%7Buser_id%7D~1view%2Fpost): the former needs a `team_id` while the latter does not, either in the URL or in the payload. (Because each channel has a unique ID and because each channel belongs at most one team, this will always be unambiguous.) And I know for certain that the async helper functions date back to APIv3 days, so I think it made sense at the time to fetch the team ID ahead-of-time and pass it everywhere instead to save needing that extra line in almost every async callback, but the new endpoints made that not as necessary.

jtdaugherty:
Okay, thanks for that detail.

jtdaugherty:
Also: for future reference, we would rather merge these changes into `develop`, not `master`.

PR #549, opened 2020-01-11
Allow submodules be cloned without a GitHub SSH key

SuperSandro2000:
I am building matterhorn in a Docker Image and would like to do that without adding a ssh key to a github account.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "submodules/aspell-pipe"]
 	path = submodules/aspell-pipe
-	url = git@github.com:matterhorn-chat/aspell-pipe.git
+	url = https://github.com/matterhorn-chat/aspell-pipe.git
 [submodule "submodules/mattermost-api-qc"]
 	path = submodules/mattermost-api-qc
-	url = git@github.com:matterhorn-chat/mattermost-api-qc.git
+	url = https://github.com/matterhorn-chat/mattermost-api-qc.git
 [submodule "submodules/mattermost-api"]
 	path = submodules/mattermost-api
-	url = git@github.com:matterhorn-chat/mattermost-api.git
+	url = https://github.com/matterhorn-chat/mattermost-api.git


jtdaugherty:
Thanks for the patch! I'd prefer not to make this change. Instead, I recommend you consider adding this step to your build process prior to initializing the submodules:
```
sed -i -e /url/s,git@github.com:,https://github.com/, .gitmodules
```

SuperSandro2000:
I mean this prevents cloning this repo without an GitHub and a tied ssh key to it.

jtdaugherty:
The repository can be cloned without SSH access, but I would agree that our current submodules can only be cloned with SSH access. The command I provided above would modify the `gitmodules` file to make it possible to fetch the submodules with HTTPS access instead.

sputnick1124:
An elegant solution to the problem could be to replace all of the absolute paths to submodules with relative ones which will inherit whatever protocol was used to clone the parent repo. This ought to avoid the issue on both sides of the equation. Internal/CI builds should complete using SSH keys, external interlopers (like myself) will not be immediately put off by not even being able to immediately clone the repository and its dependencies from a computer where they have not set up SSH authentication (like the one I'm on).

```
[submodule "submodules/mattermost-api-qc"]
        path = submodules/mattermost-api-qc
        url = ../mattermost-api-qc
[submodule "submodules/mattermost-api"]
        path = submodules/mattermost-api
        url = ../mattermost-api
[submodule "submodules/aspell-pipe"]
        path = submodules/aspell-pipe
        url = ../aspell-pipe
```

jtdaugherty:
I investigated this option a while back, and while this option does do what you describe, it breaks local clones (`git clone matterhorn other-matterhorn`).

sputnick1124:
Not sure I follow what you mean about local clones.

jtdaugherty:
I mean cloning a clone locally, i.e.,
```
$ git clone ssh://some/repo
$ git clone repo repo-copy
```
After those commands, `repo/` is the initial clone and `repo-copy/` is another git clone with `repo/` as its origin.

sputnick1124:
Ah yes I see that now. That's not something I find myself doing very often,
so wasn't considered in my suggestion.

sputnick1124:
It's been bugging me, and my curiosity is forcing me to ask: what use case(s) do you have which necessitates cloning repositories locally?

I was once in a position where I had to use local clones, but it was only a temporary situation until my organization was able to set up a git server on our internal infrastructure. 

As a side note, in that temporary situation, relative submodule paths actually *fixed* some of the problems I was facing with being able to recursive clone locally and via SSH.

kquick:
Speaking for myself, I very frequently locally clone repos.  As an example, I may be working on a feature and discover a bug I wish to work on separately, or a refactoring I would like to try before continuing to work on the feature.  Local cloning allows me to independently address these tasks and:
  *  is much more efficient (doesn't have to go over the network to the server)
  * can be used in offline conditions
  * doesn't publish branches or work to the public server that I'm not ready to publish.

Notably, git as a distributed VCS tool is designed to be used and is entirely useable without a forge like Github or Gitlab (and it predates the development of the forges).  The forges provide a very convenient way to share work and coordinate with other people and remove the need for self-hosting, but they are still an optional portion of the VCS model.

jtdaugherty:
@sputnick1124 I don't use local clones as much as @kquick, but I cited that example because I do find local clones useful when doing scripting (e.g. for releases), and I don't want to break local clones to deal with this `git` shortcoming when the alternative is (to me) less unpleasant.

sputnick1124:
@kquick, that makes sense.

@jtdaugherty please don't mistake my curiosity for criticism. I totally understand your decision to keep the current configuration and am not trying to convince you otherwise. I'm just always interested to learn about other peoples' workflows and how they affect their processes.

Thanks for enduring my questions!

jtdaugherty:
Thanks @sputnick1124 - I didn't interpret your comment that way, but I just wanted to be clear about my preference.

PR #573, opened 2020-03-16
Added '--' to notify script

geirskjo:
to avoid argument error "unknown option" for messages beginning with a
"-"


jtdaugherty:
Thank you!

Issue #574, opened 2020-03-16
Consider consolidated server URL input in login form


jtdaugherty:
From a comment by @glguy:

An alternative design [to the current login form's individual inputs for URL components] would be to not separate the host and path portion but to allow the user to specify them all in one go. This has the advantage that the user can basically paste what he sees in a web browser into matterhorn. I'm imagining this UI:

```
          ┌──────────────────────────────────────────────────────────┐          
          │Please enter your Mattermost credentials to log in.       │          
          │                                                          │          
          │Server:           mattermost.example.com:443/path/part    │          
          │                                                          │
          │Username:         emertens                                │          
          │                                                          │          
          │Password:                                                 │          
          │                                                          │          
          │          Press Enter to log in or Esc to exit.           │          
          └──────────────────────────────────────────────────────────┘
```

In the most common case where we use the defaults a user might only enter

```
          │Server:           mattermost.example.com                  │          
```

kquick:
I think we might still need to separate the elements (via urlparse) to establish websocket as well as direct server access (just a hunch, I haven't looked specifically).  It would be worth spending some archaeological effort determining why this approach wasn't taken when the login process was initially designed.

If we were to try to use urlparse, we would probably need to either synthesize the protocol specification prefix (e.g. `https://`) or require the user to input that part as well. 

Alternatively we could internally parse a subset like the above, but we would want to have either or both of (a) good documentation about what is expected (probably right there on the login screen) and (b) resiliency in the parser for the user providing extra elements like the protocol.

jtdaugherty:
To review the listed advantages above:
* A single field makes entry easier when only the server's hostname is known.
* A single field makes entry easier even for the full URL case, scheme and path and all.

Some things we need to consider:
* A URL entered with no scheme, port, or path should default to `https`, `443`, and `/` respectively.
* If the scheme and/or port are provided explicitly, they should be honored: so `https://foo:123/` or `http://foo:443/` should be taken literally.
* Schemes other than `http` and `https` should be rejected.
* To be safe, the input handling should trim anything after the URL path (query string, fragment, etc.).

To @kquick's comment, there is probably not going to be any evidence anywhere as to why this was not the original approach. (It was not documented or even discussed explicitly, IIRC.)

[...]

PR #577, opened 2020-03-16
consolidated login screen

glguy:
Discussion on #574 

This preserves the existing configuration file format. The input field is validated using a URI parser and query and fragment parts are rejected

Examples of supported inputs:

- `mattermost.example.com`
- `mattermost.example.com:4343`
- `mattermost.example.com/one/two`
- `mattermost.example.com:4343/one/two`


jtdaugherty:
That appears to not support inputs like `http://server.com/`. I guess to support that, `ConnectionInfo` would need to get a field to store whether HTTPS is intended.

Issue #515, opened 2019-08-16
Support authentication via Personal Access Token


setpill:
This would allow users to use matterhorn on mattermost instances where the only allowed auth method is gitlab auth by extracting the `MMAUTHTOKEN` from their mattermost instance cookie after logging in through the webinterface.

setpill:
See also https://github.com/42wim/matterircd/issues/29

Atlantic777:
I'm also interested in this feature and would like to contribute. I'll have a closer look at this. I need either the tokens or something more to properly support OAuth.

jtdaugherty:
From my conversation today with Harrison Healey, one of the Mattermost developers:
```
[10:49] jtdaugherty: @harrison if we wanted to support Oauth Personal Access Tokens in Matterhorn, what's the user experience for getting those?
                     Is there a browser in the loop? Is it possible to get those from the user without a browser in the loop during
                     authentication?
[11:03] │ harrison: That's a good question. I'll have to look to see how you get those
[11:04] │ harrison: As long as they're enabled on the server, I assume it's just a simple API call to create one
[11:06] │ harrison: Yeah, there's just a simple API to create them as long as they're enabled on the server. Like here's the API to create one:
        │           https://api.mattermost.com/#tag/users%2Fpaths%2F~1users~1%7Buser_id%7D~1tokens%2Fpost
[11:07] │ harrison: Normally, the user can create them from the Security > Personal Access Tokens section of the account settings section of the
        │           web app when they're enabled and the user has permission to be able to do it
[11:52] │ jtdaugherty: Oh, well if there's a UI for it, then that's all we need - we'd just direct the user to that UI, and then they'd provide
        │              the access token to Matterhorn in an auth UI or configuration file.
[11:53] │ jtdaugherty: My main concern was whether we'd have to modify Matterhorn to participate in any OAuth protocol flow.
[11:53] │ jtdaugherty: But if it just amounts to a password we include in an auth request, then that's easy.
```

jtdaugherty:
Also, thanks to all who are willing to help out with this! If you decide to contribute, please take a look at this and keep it in mind:

https://github.com/matterhorn-chat/matterhorn/blob/master/PRACTICES.md

In particular, before supporting this feature we need to work out (in the comments on this issue or on the wiki) the user experience that will be provided. For example, a starting-point proposal (assuming the above conversation about access tokens is right) is:

[...]

PR #581, opened 2020-03-24
Enable token authentication

ftilde:
This is my attempt at resolving #515.

Since in the current situation we are using mattermost more and more at work, I just went ahead and implemented support for token authentication without more discussion in the issue. The current solution works for me for now, but of course I'm very much open to discussion on the specifics.

As far as I know, and as pointed out in #515, OAUTH and personal access tokens only differ in the way they are obtained and how long they are valid. So this approach would work for both. (Although the instance I use does not offer an option to obtain personal access tokens and I have only tested this with OAUTH tokens obtained from the cookies of my web browser. Somebody else should probably test this as well.)

In contrast to what was discussed in #515, I have only implemented the `tokencmd` config option for simplicity. As discussed there another option for reading the token from a file is (1.) questionable from a security point of view and (2.) can just be emulated using `tokencmd: cat /path/to/token/file`.

In terms of code I should point out that I have removed the function `getCredentials` instead of changing it as it wasn't used anywhere.

Finally, I have to admit that the last time I've really used haskell was in an undergraduate course at university so... If you have any requests for style changes I'm not offended in any way.

### DIFF
--- a/docs/sample-config.ini
+++ b/docs/sample-config.ini
@@ -24,8 +24,8 @@
 #
 # port: 443
 
-# Password command. Optional. If this and the password option are both
-# missing or give the wrong password, you'll be prompted on startup.
+# Password command. Optional. If this, the password option and token command
+# are all missing or give the wrong password, you'll be prompted on startup.
 #
 # You can also just use
 #
@@ -64,6 +64,19 @@
 # You can then set:
 #
 # passcmd: secret-tool lookup matterhorn password
+
+# Token command. Optional. If this is specified, the provided command will be
+# executed to obtain an OAUTH or personal access token that can be used instead
+# of username and password for authentication with the mattermost server. The
+# token command takes precedence over username and password configuration
+# options, i.e., when a token is specified username and password options are
+# ignored.
+#
+# Otherwise the token command functions just like the password command. Hence,
+# if you have obtained a personal access token and stored it in your keychain,
+# you can use it like this:
+#
+# tokencmd: secret-tool lookup matterhorn token
 
 # This optional setting controls how the client displays times. If it's
 # absent, you get the default behavior ("%R"). If it's present but
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -5,7 +5,6 @@ module Config
   ( Config(..)
   , PasswordSource(..)
   , findConfig
-  , getCredentials
   , defaultConfig
   )
 where
@@ -107,6 +106,8 @@ fromIni = do
     configPass <- (Just . PasswordCommand <$> field "passcmd") <|>
                   (Just . PasswordString  <$> field "pass") <|>
                   pure Nothing
+    configToken <- (Just . TokenCommand  <$> field "tokencmd") <|>
+                  pure Nothing
     configUnsafeUseHTTP <-
       fieldFlagDef "unsafeUseUnauthenticatedConnection" False
     configValidateServerCertificate <-
@@ -183,6 +184,7 @@ defaultConfig =
            , configTeam                        = Nothing
            , configPort                        = defaultPort
            , configPass                        = Nothing
+           , configToken                       = Nothing
            , configTimeFormat                  = Nothing
            , configDateFormat                  = Nothing
            , configTheme                       = Nothing
@@ -285,23 +287,16 @@ getConfig fp = do
                 Just (PasswordString pass) -> return $ Just pass
                 Nothing -> return Nothing
 
+            actualToken <- case configToken conf of
+                Just (TokenCommand cmdString) -> do
+                    let (cmd:rest) = T.unpack <$> T.words cmdString
+                    output <- convertIOException (readProcess cmd rest "") `catchE`
+                              (\e -> throwE $ "Could not execute token command: " <> e)
+                    return $ Just $ T.pack (takeWhile (/= '\n') output)
+                Just (TokenString pass) -> error $ "getConfig: Token was provided as plain text. This is a bug!"
+                Nothing -> return Nothing
+
             return conf { configPass = PasswordString <$> actualPass
+                        , configToken = TokenString <$> actualToken
                         , configAbsPath = Just absPath
                         }
-
--- | Returns the hostname, username, and password from the config. Only
--- returns Just if all three have been provided. The idea is that if
--- this returns Nothing, we're missing at least some of these values.
-getCredentials :: Config -> Maybe ConnectionInfo
-getCredentials config = do
-    pass <- configPass config
-    passStr <- case pass of
-        PasswordString p -> return p
-        PasswordCommand _ ->
-            error $ "BUG: unexpected credentials state: " <>
-                    show (configPass config)
-
-    ConnectionInfo <$> configHost config
-                   <*> (pure $ configPort config)
-                   <*> configUser config
-                   <*> (pure passStr)
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -293,7 +293,7 @@ getConfig fp = do
                     output <- convertIOException (readProcess cmd rest "") `catchE`
                               (\e -> throwE $ "Could not execute token command: " <> e)
                     return $ Just $ T.pack (takeWhile (/= '\n') output)
-                Just (TokenString pass) -> error $ "getConfig: Token was provided as plain text. This is a bug!"
+                Just (TokenString _) -> error $ "getConfig: Token was provided as plain text. This is a bug!"
                 Nothing -> return Nothing
 
             return conf { configPass = PasswordString <$> actualPass
--- a/src/Login.hs
+++ b/src/Login.hs
@@ -66,21 +66,25 @@ import           Control.Concurrent ( forkIO, threadDelay )
 import           Control.Exception ( catch )
 import           System.IO.Error ( catchIOError )
 import qualified Data.Text as T
+import           Data.Text (unpack)
 import           Graphics.Vty hiding (mkVty)
 import           Lens.Micro.Platform ( (.~), Lens', makeLenses )
 import qualified System.IO.Error as Err
 
 import           Network.Mattermost ( ConnectionData )
-import           Network.Mattermost.Types ( Session, User, Login(..), ConnectionPoolConfig(..)
-                                          , initConnectionData, ConnectionType(..) )
+import           Network.Mattermost.Types ( Session(..), User, UserParam(..), Login(..), ConnectionPoolConfig(..)
+                                          , initConnectionData, ConnectionType(..))
+import           Network.Mattermost.Types.Internal ( Token(..) )
 import           Network.Mattermost.Exceptions ( LoginFailureException(..) )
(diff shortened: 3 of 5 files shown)


jtdaugherty:
Thank you for the patch! I will set aside some time to take a look at this in the next few days.

jtdaugherty:
I spent time today taking a look at this. After reading over [the Mattermost page on OAuth](https://docs.mattermost.com/developer/oauth-2-0-applications.html), I'm curious: as you use Matterhorn with this feature, on behalf of what OAuth application are you getting a token? You mentioned getting it from your browser cookie, which sounds to me like you're pairing up Mattermost with some browser-based OAuth app, then swiping the cookie for that and re-using it for Matterhorn. Is that right?

jtdaugherty:
As far as the patch itself is concerned, I have to go ahead and point out that it cannot be merged for a few reasons.

* The patch is against `master`, whereas it needs to be made against `develop` instead, and `develop` has seen quite a few changes to the Login UI lately. Those changes would conflict directly with this patch. On seeing this I realized we didn't mention this important contribution guideline in the README, [so I added it](https://github.com/matterhorn-chat/matterhorn/commit/510d48346f90180e6134ba9cb2c47ac760a93519). Sorry about that!
* I think the design needs to be adjusted. The patch does not provide a login UI field for the token, and I think that's something we'd want to add if we start supporting token-based authentication. In addition, there are some issues with how the patch is implemented that lead to code duplication and avoidable bug checks. If the `develop` conflict weren't the case, I'd probably go ahead and comment on those in an effort to move this patch forward. I can still do that if you would find it helpful, but since there is more work/thinking to be done, I'm holding off on that for now.
* My biggest concern with supporting any token-based authentication is that we need to have a good understanding of the user experience that we are asking people to participate in, in order to provide Matterhorn with a token. That's something we'll need to document. And since OAuth assumes an HTTP/redirect-based application, that presents some usability challenges when your app is a terminal application. We need to understand that impedance mismatch really well so we can document the process clearly, and we also need to understand the security implications of supporting token-based access.

With all that said, I'm super glad that you were able to modify Matterhorn to meet your needs, I'm glad you are using Matterhorn at all, and this patch is helpful to us to see how this might be done. So, thank you!

To me, the next step depends on the answer to my question above about how to get the token. I'm concerned that getting a token from a cookie (i.e. for a different OAuth application) essentially breaks the OAuth model. Based on whether that's true, I'd want to think through other ways for Matterhorn could work. (Hosting its own web server comes to mind.)

jtdaugherty:
(On the topic of Matterhorn hosting a server: that idea comes from [posts like this](https://developer.okta.com/blog/2018/07/16/oauth-2-command-line).)

ftilde:
I'm very sorry about the confusion. This is entirely my fault. I've made the initial very hacky version of this PR some time ago and only sat down this week to clean it up, but did not really re-read all of the available material. Let me try to clarify this a bit:

This really PR (primarily) adds support for login via personal access tokens. It so happens that this also enables login in using the `MMAUTHTOKEN` (obtained from the browser after login via the web-interface) which I believe is set when using the gitlab auth method (where mattermost acts as a OAUTH-client I think? Not sure about that). (This is the only login option offered by the mattermost instance I use for work, so there is really no other way for me to use matterhorn there.)

Despite what I've written above, this PR does not enable authentication via OAUTH with the mattermost instance as an authorization server itself. As you have pointed out and as is documented in the linked article, this would involve a webserver acting as the OAUTH-client, and is probably far more work that what I've done here.

With that out of the way in response to the other points you made:

> The patch is against master, whereas it needs to be made against develop instead, and develop has seen quite a few changes to the Login UI lately. Those changes would conflict directly with this patch. On seeing this I realized we didn't mention this important contribution guideline in the README, so I added it. Sorry about that!

Oh well, I guess I could have noticed that issue as well. I'm willing to rework the PR so that it is aimed at `develop` if you think that the approach is otherwise ok (or will be ok when some more changes are made).

> I think the design needs to be adjusted. The patch does not provide a login UI field for the token, and I think that's something we'd want to add if we start supporting token-based authentication. In addition, there are some issues with how the patch is implemented that lead to code duplication and avoidable bug checks. If the develop conflict weren't the case, I'd probably go ahead and comment on those in an effort to move this patch forward. I can still do that if you would find it helpful, but since there is more work/thinking to be done, I'm holding off on that for now.

The fact that there is no UI for token-based authentication was (more or less) a deliberate design decision. In contrast to a password (which may just remember if you are not using a password manager) I don't see why a user would type out their personal access token by hand into the login TUI. As this is more or less a more advanced feature I thought it would be fine if it were only available via the config file. With that said, I'm of course open to other opinions on the topic.

> My biggest concern with supporting any token-based authentication is that we need to have a good understanding of the user experience that we are asking people to participate in, in order to provide Matterhorn with a token. That's something we'll need to document. And since OAuth assumes an HTTP/redirect-based application, that presents some usability challenges when your app is a terminal application. We need to understand that impedance mismatch really well so we can document the process clearly, and we also need to understand the security implications of supporting token-based access.

For personal access tokens the user experience would probably be (right now): Obtain a personal access token via the web interface of the instance, store it in your password manager and add the corresponding `tokencmd` to your config. (I realize that your comment is mostly directed towards OAUTH authentication due to my original message, of course.)

jtdaugherty:
> This PR (primarily) adds support for login via personal access tokens. It so happens that this also enables logging in using the MMAUTHTOKEN (obtained from the browser after login via the web-interface) which I believe is set when using the gitlab auth method (where mattermost acts as a OAUTH-client I think? Not sure about that).

Okay, thanks for the clarification! To clarify: is `MMAUTHTOKEN` the cookie name?

> In contrast to a password (which may just remember if you are not using a password manager) I don't see why a user would type out their personal access token by hand into the login TUI. As this is more or less a more advanced feature I thought it would be fine if it were only available via the config file.

That's definitely reasonable to me. The reason why I would like it to be included in the login UI is because the login UI is intended mostly for first-time users: they don't know about the config file yet and just want to get logged in. Having to set up a config before being able to try out the client is, to me, too high-friction. The login UI is there to allow them to do that interactively, so I'd like it to support all auth methods. I would then imagine that our docs would say "consider adding your access token to your keychain and then set up `tokencmd`." Even the login UI could provide that recommendation if the user inputs a token.

jtdaugherty:
Another thing to clarify: judging by my reading of the Mattermost docs, "personal access tokens" are different from the token you got from your GitLab cookie, but it sounds like you are saying that it doesn't matter which kind you have because you can use either one in this context. Have you tested both? (It sounds like you have necessarily tested the GitLab version, but I wonder if you have created a personal access token and tried that, too.)

jtdaugherty:
I just tested this implementation with a Personal Access Token and it did indeed work.

jtdaugherty:
FYI I'm now in the process of writing up a new patch based on this one in the `feature/access-tokens` branch.

ftilde:
> Okay, thanks for the clarification! To clarify: is MMAUTHTOKEN the cookie name?

Yes, indeed. To me it looks like it is "just" the session cookie, which is also set when logging in with username and password. I just tested this with a test account at `https://gitlab.mattermost.com`: Log in via webinterface, grab cookie, and login with matterhorn using `tokencmd: echo COOKIE`.

Quoting from the [documentation](https://api.mattermost.com/#tag/authentication):

> Using personal access tokens is very similar to using a session token. The only real difference is that session tokens will expire, while personal access tokens will live until they are manually revoked by the user or an admin.

This explains why using another token instead of the personal access token actually works.

> I just tested this implementation with a Personal Access Token and it did indeed work.

No, I have not tested Personal Access Tokens myself, as I do not have access to an instance that offers them. I spent some time trying to spin up a test instance locally, but did not get that far. Glad that you could confirm that it works!

> That's definitely reasonable to me. The reason why I would like it to be included in the login UI is because the login UI is intended mostly for first-time users: they don't know about the config file yet and just want to get logged in. Having to set up a config before being able to try out the client is, to me, too high-friction. The login UI is there to allow them to do that interactively, so I'd like it to support all auth methods. I would then imagine that our docs would say "consider adding your access token to your keychain and then set up tokencmd." Even the login UI could provide that recommendation if the user inputs a token.

Fair enough! I definitely see how that would be beneficial.

> FYI I'm now in the process of writing up a new patch based on this one in the feature/access-tokens branch.

Awesome! I probably won't be available over the weekend, but would be able to help some time next week, if desired.

jtdaugherty:
[The branch](https://github.com/matterhorn-chat/matterhorn/tree/feature/access-tokens) is ready for testing. @ftilde if you want to try it out, that would be great; it should be compatible with your current configuration.

ftilde:
It works both with my existing configuration and when pasting the token in the TUI. Great!

jtdaugherty:
Fantastic! I'm going to write up some documentation for users on how to get both types of token and then I think it'll be merge-ready.

jtdaugherty:
Another clarification, @ftilde - is the `MMAUTHTOKEN` cookie on the domain for the Mattermost web client?

ftilde:
> Another clarification, @ftilde - is the MMAUTHTOKEN cookie on the domain for the Mattermost web client?

That is correct!

jtdaugherty:
The branch now has [a revamped README section on authentication](https://github.com/matterhorn-chat/matterhorn/tree/feature/access-tokens#authentication). Please check it out!

jtdaugherty:
Since we seem to have concluded this, I'm going to close this PR. Thanks for working on this and thanks for all the details above; that helped me tremendously.

ftilde:
@jtdaugherty Sorry for the late reply. I like the updated readme. One comment though:

It looks like the MMAUTHTOKEN-cookie ist *just* the [session cookie](https://github.com/mattermost/mattermost-server/blob/a63684fcb5e3ba7b7522b35c29a4cb27779ba823/model/session.go#L13) of the web client and doesn't have anything to do with OAUTH in itself.
In particular, an OAUTH token obtained when communicating with the mattermost server acting as the oauth-authorization server would be something else entirely and *not* be usable in place of the personal access token. I suspect so at least, because I've read that OAUTH tokens are supposed to be very short-lived. At the very least that would make them very inconvenient to use with `tokencmd`.

I think it would make sense to stress the gitlab-authentication aspect in the third paragraph of the authentication section. Something like:


## GitLab authentication

If your server does not support username/password or personal access token authentication (e.g. when GitLab authentication is the only option), follow these steps to authenticate with the session  token of the web client:

* Authenticate to GitLab using the Mattermost web client.
* Once your browser has returned to the Mattermost interace, obtain the value of the MMAUTHTOKEN browser cookie.
* Start Matterhorn and enter the value of the MMAUTHTOKEN token in the login interface field labeled Access token: or add the token to the system keychain as described above.

Please note that the session token of the web client will expire (in contrast to a personal access token), so you will have to repeat the above procedure from time to time.

jtdaugherty:
Thanks! I just clarified this in 8865b64d9239d0f39b4e61f6e4b5dcdbf0848a53.

jtdaugherty:
FYI the work done on this topic was included in 50200.8.0 which I just released.

PR #601, opened 2020-04-15
State.Editing: Bind moving one word left and right to C-arrow keys

### LINKED ISSUE #599: Support C-Left and C-Right for moving over words
As a new user to matterhorn I encountered two small "paper cut" issues when editing messages.

1. Home did not skip to the beginning of the line: already discussed in #258 and trivially fixable by unbinding `scroll-top`. But just to note, it did not work for me even in multi-line mode: a regression or was this never implemented?
2. Moving over words with C-Left and C-Right did not work. Basically the same behavior as provided by M-b and M-f, just alternative keybindings. The former are widely supported by other terminals and editors and are more comfortable on my layout.

It would be nice to add support for C-Left and C-Right.

thsnr:
fixes #599

### DIFF
--- a/src/State/Editing.hs
+++ b/src/State/Editing.hs
@@ -152,8 +152,14 @@ editingKeybindings =
   , kb "Move one word to the right"
     (EvKey (KChar 'f') [MMeta]) $ do
     csEditState.cedEditor %= applyEdit Z.moveWordRight
+  , kb "Move one word to the right"
+    (EvKey KRight [MCtrl]) $ do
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks for the patch!

Unfortunately I'd rather not merge it, because I would like to take a different approach to address that ticket. I'm happy to discuss it there, but I need to do a bit more investigation to check on some things.

thsnr:
No problem, that was a very simplistic approach to the problem and I will happily wait if you have something better in mind.

PR #612, opened 2020-04-29
Use correct link to community server

immindich:
The README currently links to [pre-release.mattermost.com](pre-release.mattermost.com), which seems to be broken. The Mattermost developers [changed their documentation](https://github.com/mattermost/mattermost-server/pull/13379) a few months ago to link to [community.mattermost.com](community.mattermost.com), so we should probably do that too.


jtdaugherty:
Thank you! I had run into that name change but I didn't realize we had a link to it.

PR #637, opened 2020-07-08
Correctly handle muted channels

immindich:
This PR allows Matterhorn to properly handle muted channels. When a channel is muted, Matterhorn behaves as follows:
- Mentions will not trigger notifications. If there are unread mentions, the channel is highlighted like an unmuted channel with unread messages and no mentions, although the message count is still displayed.
- If there are no mentions, the channel is treated as having no unread messages. A muted channel with unread messages will have `(m)` in the place where the mention counter normally is.
- Since DMs are treated similarly to mentions, a muted DM channel will behave similarly to a regular channel (messages appear as regular unreads and do not cause notifications).

Muting a channel can be done with `/mute`, which is already a server command.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -481,8 +481,10 @@ hasUnread' :: ClientChannel -> Bool
 hasUnread' chan = fromMaybe False $ do
     let info = _ccInfo chan
     lastViewTime <- _cdViewed info
-    return $ ((_cdUpdated info) > lastViewTime) ||
-             (isJust $ _cdEditedMessageThreshold info)
+    return $ _cdMentionCount info > 0 ||
+             (not (isMuted chan) &&
+              ((_cdUpdated info) > lastViewTime) ||
+              (isJust $ _cdEditedMessageThreshold info))
 
 mkChannelZipperList :: UTCTime
                     -> Config
--- a/src/Draw/ChannelList.hs
+++ b/src/Draw/ChannelList.hs
@@ -47,6 +47,7 @@ data ChannelListEntryData =
                          , entryIsRecent    :: Bool
                          , entryIsReturn    :: Bool
                          , entryIsCurrent   :: Bool
+                         , entryIsMuted     :: Bool
                          , entryUserStatus  :: Maybe UserStatus
                          }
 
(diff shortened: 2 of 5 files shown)


src/Types.hs
```diff
@@ -481,8 +481,10 @@ hasUnread' :: ClientChannel -> Bool
 hasUnread' chan = fromMaybe False $ do
     let info = _ccInfo chan
     lastViewTime <- _cdViewed info
-    return $ ((_cdUpdated info) > lastViewTime) ||
-             (isJust $ _cdEditedMessageThreshold info)
+    return $ _cdMentionCount info > 0 ||
```
jtdaugherty, line 484: I'm curious why the `_cdMentionCount` check occurs even if the channel is muted. Is that how the web client behaves?

jtdaugherty:
Thank you for this work, @immindich! I left a comment. Otherwise I have just one question: what does the web client do if:
* a channel A is muted
* channel A receives new messages
* channel A is unmuted

Does channel A appear as having unread messages in the sidebar? That is, unread messages relative to the point at which it was muted? Or does it appear as having no unread activity? If the latter, we should probably also update the channel view timestamp when unmuting the channel with `State.Channels.updateViewedChan` since that will reset the unread status in the server metadata (and thus our sidebar logic).

src/Types.hs
immindich: The web client treats mentions in a muted channel as unread messages. If you use keyboard shortcuts to jump to the next channel with unread messages, this includes muted channels with mentions. These channels are displayed the same way as channels with mentions normally are, except darker. Here's what this looks like for me (testing999 is muted, but Town Square isn't):
Screen Shot 2020-07-08 at 9 51 44 AM

In Matterhorn on this branch, testing999 is highlighted as if it has normal unread messages, but the mention count is still displayed.
Screen Shot 2020-07-08 at 9 55 19 AM

jtdaugherty: Okay, thanks!

immindich:
The web client considers those messages unread. However, it is unlikely that a user will actually notice this, since you need to be viewing a channel to unmute it.

jtdaugherty:
Okay, that makes sense. Thank you!

PR #650, opened 2020-08-11
Add interface for editing channel notification preferences

immindich:
This PR adds a command `/notify-prefs` which allows the user to change the notification preferences for the current channel. A few notes:

- This does not allow the user to change the global notification preferences.

- Matterhorn only checks to see if the user is attempting to save or cancel and passes all other keypresses to `handleFormEvent`. This means that the keys for changing the form state cannot be rebound. If we want to make them rebindable, I guess I could have Matterhorn handle all the key events and update the form state directly.

- I only added the options which could be changed from the web client. The Mattermost API also has a per-channel setting for email notifications.

- Changing the notification preferences is disabled on individual DM channels, just like in the web client. You can still use `/mute` to mute a DM channel.

### DIFF
--- a/matterhorn.cabal
+++ b/matterhorn.cabal
@@ -111,6 +111,7 @@ executable matterhorn
                        Events.LeaveChannelConfirm
                        Events.DeleteChannelConfirm
                        Events.TabbedWindow
+                       Events.EditNotifyPrefs
                        Windows.ViewMessage
                        HelpTopics
                        KeyMap
--- a/matterhorn.cabal
+++ b/matterhorn.cabal
@@ -59,6 +59,7 @@ executable matterhorn
                        State.Logging
                        State.Messages
                        State.MessageSelect
+                       State.NotifyPrefs
                        State.PostListOverlay
                        State.ThemeListOverlay
                        State.ReactionEmojiListOverlay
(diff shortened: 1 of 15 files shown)


jtdaugherty:
> This does not allow the user to change the global notification preferences.

Thanks - we can add that later.

> Matterhorn only checks to see if the user is attempting to save or cancel and passes all other keypresses to handleFormEvent. This means that the keys for changing the form state cannot be rebound. If we want to make them rebindable, I guess I could have Matterhorn handle all the key events and update the form state directly.

I'm okay with not making these rebindable for now. I can't really see a ton of value in being able to rebind keys that are likely to be reasonable for 99.9% of people. If we get any requests to make them rebindable, we can reconsider. For now, the fact that they're documented in the UI itself also makes it okay (to me) to not mention them in the generated keybinding help.

> I only added the options which could be changed from the web client. The Mattermost API also has a per-channel setting for email notifications.

Okay, good to know.

> Changing the notification preferences is disabled on individual DM channels, just like in the web client. You can still use /mute to mute a DM channel.

Makes sense.

Thank you for the details!

jtdaugherty:
The patches look good to me. Thanks so much for your work on this!

PR #659, opened 2020-08-27
Use https urls for submodules

popey:
When building matterhorn behind a proxy, using git@ urls for submodules causes builds to fail. Using https urls for submodules fixes this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "submodules/mattermost-api-qc"]
 	path = submodules/mattermost-api-qc
-	url = git@github.com:matterhorn-chat/mattermost-api-qc.git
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Can you help me understand why you need to do source builds?

popey:
I maintain a [snap](https://snapcraft.io/matterhorn) of matterhorn. Currently it builds by simply re-packing the existing built binaries. However, we have a build system which can do builds on any commit to master, so adventurous users can try out the latest code drops by subscribing to the 'edge' channel in the snap store. The build happens inside the launchpad.net datacentre, and those machines won't have access to the git@ accessible submodules.

jtdaugherty:
I see, thanks! In that case I recommend modifying `.gitmodules` at build time with `sed` as described in `mkrelease.sh`.

Issue #228, opened 2017-05-08
Support for 2FA authentication


madchap:
Hi,

I know it is mentioned in the README.md that only user/pass is currently supported, but I could not find any outstanding issue asking for 2FA support.

Thank you.
Best,
fred

jtdaugherty:
Thanks! We definitely need to take a look at this at some point. It is not on our current development plan (which runs through August) but we'd definitely be open to working with contributors if anyone wants to take a look at this.

madchap:
Hi,

Just pinging... everytime I look back to Matterhorn, I am like "why am I not using this?". And it strikes me again :-( I hope you can get this prioritized somehow :+1:

jtdaugherty:
@madchap - thanks for the ping. It's good to know that this would make Matterhorn useful to you!

To figure out when we might tackle this, we first need to investigate what would be required to do it, and in particular, whether it is even feasible (in case something about how it is implemented depends on the web client). We could at least make that investigation a priority in the coming months, but only after that's completed could we say anything definite about implementation. If someone else (such as yourself) wanted to do that, here's what we'd want to know:

* What is the protocol workflow for 2FA? What API requests need to be made between which entities, in what order?
* What are the user interface requirements to effectively support 2FA?
* Does the web browser need to be involved at all?
* For implementation, are there already helper libraries or would we have to implement new ones?

madchap:
Pinging my friend @benoitknecht here, just in case :-P

dylangerdaly:
If putting 2FA is a little too hard, would it be useful to allow for session 'injection'

Where you can specify the hostname, username and cookies and just log into an already active session?

jtdaugherty:
@dylangerdaly no, that would not be advisable.

jtdaugherty:
This has been migrated to the wiki: https://github.com/matterhorn-chat/matterhorn/wiki/Feature-Backlog

xfalcox:
I got MFA working with a very ugly hack (long time without doing haskell)

https://github.com/xfalcox/matterhorn/commit/7d1aceeedc72890d2f01ce28cc61f2039e218ea7

kquick:
Thanks for the update, @xfalcox !  Can you tell me why you think this is an ugly hack?  There are a few minor comments, but it looks like a pretty reasonable implementation at first glance.

[...]

PR #663, opened 2020-09-03
Support login with 2FA token

msm-code:
**I don't think this should be merged as it is**.

So I've hacked together this code, to keep using matterhorn. This is how login form looks after my changes:

ss

This works for me, but is ugly when user *don't* need 2FA token to login.

I'm open for improving this PR, but looking for feedback first.

See also my post at #228.

Note: this depends on https://github.com/matterhorn-chat/mattermost-api/pull/56.

### DIFF
--- a/src/Matterhorn/Login.hs
+++ b/src/Matterhorn/Login.hs
@@ -48,7 +48,7 @@
 -- fails before the timer fires, we just resume normal operation and
 -- show the login form so the user can intervene.
 module Matterhorn.Login
-  ( LoginAttempt(..)
+  ( LoginData(..)
   , interactiveGetLoginSession
   )
 where
@@ -110,6 +110,12 @@ data LoginAttempt =
     -- ^ The attempt succeeded, but additional MFA token is required.
     | AttemptSucceeded ConnectionInfo ConnectionData Session User (Maybe Text) --team
     -- ^ The attempt succeeded.
+
+-- | The result of a successfull login attempt.
+data LoginData =
+    LoginSuccess ConnectionData Session User (Maybe Text) --team
+    -- ^ Data associated with the new logged-in session.
+
 
 -- | The state of the login interface: whether a login attempt is
 -- currently in progress.
@@ -313,7 +319,7 @@ interactiveGetLoginSession :: Vty
                            -- info provided here is fully populated, an
                            -- initial connection attempt is made without
                            -- first getting the user to hit Enter.
-                           -> IO (Maybe LoginAttempt, Vty)
+                           -> IO (Maybe LoginData, Vty)
 interactiveGetLoginSession vty mkVty setLogger logMgr initialConfig = do
     requestChan <- newBChan 10
     respChan <- newBChan 10
@@ -332,7 +338,9 @@ interactiveGetLoginSession vty mkVty setLogger logMgr initialConfig = do
 
     (finalSt, finalVty) <- customMainWithVty vty mkVty (Just respChan) app startState
 
-    return (finalSt^.lastAttempt, finalVty)
+    return $ case finalSt^.lastAttempt of
+        Just (AttemptSucceeded _ cd sess user mbTeam) -> (Just $ LoginSuccess cd sess user mbTeam, finalVty)
+        _ -> (Nothing, finalVty)
 
 -- | Is the specified ConnectionInfo sufficiently populated for us to
 -- bother attempting to use it to connect?
--- a/src/Matterhorn/Login.hs
+++ b/src/Matterhorn/Login.hs
@@ -90,7 +90,6 @@ import           Matterhorn.Types ( ConnectionInfo(..)
                        , ciAccessToken, ciToken, SemEq(..)
                        )
 
-
 -- | Resource names for the login interface.
 data Name =
       Server
@@ -107,6 +106,8 @@ instance SemEq Name where
 data LoginAttempt =
     AttemptFailed AuthenticationException
     -- ^ The attempt failed with the corresponding error.
+    | MfaTokenRequired ConnectionInfo
+    -- ^ The attempt succeeded, but additional MFA token is required.
     | AttemptSucceeded ConnectionInfo ConnectionData Session User (Maybe Text) --team
     -- ^ The attempt succeeded.
 
@@ -217,6 +218,9 @@ loginWorker setLogger logMgr requestChan respChan = forever $ do
 
                           result <- convertLoginExceptions $ mmLogin cd login
                           case result of
+                              Left (MattermostServerError (MattermostError {mattermostErrorId = "mfa.validate_token.authenticate.app_error"})) -> do
+                                  doLog $ "Authenticated successfully to " <> connInfo^.ciHostname <> " but MFA token is required"
+                                  writeBChan respChan $ LoginResult $ MfaTokenRequired connInfo
                               Left e -> do
                                   doLog $ "Error authenticating to " <> connInfo^.ciHostname <> ": " <> (T.pack $ show e)
                                   writeBChan respChan $ LoginResult $ AttemptFailed e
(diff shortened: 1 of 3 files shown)


eyenx:
Hmm Thanks gonna try it out quickly.

How about just use the inputted OTP Token if it is inputted by the user. If the OTP content stays "Empty" do the normal login?

eyenx:
also a `otpcmd` for the config.ini would be nice.

eyenx:
But it works for me! :D thanks. Even for some of our colleagues are work it is usable

jtdaugherty:
@eyenx Thanks for giving this a try! You can find additional context and discussion here:

https://github.com/matterhorn-chat/matterhorn/issues/228

msm-code:
I've pushed my changes - PTAL again.

I've added a modal login window. The first stage looks as before:

image

But after a successful login, OTP window shows up:

image

msm-code:
Oh, by the way - I have no idea what I'm doing ;).

I have some experience with "academia" Haskell, but no experience with "commercial" Haskell. Many idioms in the codebase are completely foreign to me. So please do review carefully.

Known caveats:

I didn't add `MfaTokenRequired` case to Setup.hs (line 92). I *think* it's not possible for `interactiveGetLoginSession` to return `MfaTokenRequired` (only `AttemptSucceeded` or `AttemptFailed`).

jtdaugherty:
Thank you for the update! I left comments on the commits requesting a few minor changes. Overall, though, the changes look good to me and I appreciate you putting in more effort on this and updating it to gather the OTP token modally!

jtdaugherty:
A note about the `Setup.hs` change you mentioned not making: you will definitely need to add a case to handle that, because not handling it will generate a compiler warning at best, and a runtime crash at worst if we fail to handle that case. What we typically do in this kind of case is to call `error` to indicate that an unexpected/illegal situation was encountered. (Ideally we just wouldn't even have a constructor in the return type, but that would require introducing an additional type to guard against that here. I think for now it's okay to leave the types as-is and call `error`.)

So the `Setup.hs` case would be something like
```haskell
```

jtdaugherty:
If you are interested in refactoring things to avoid adding the additional case (which is admittedly a bit sloppy and something that usually tells us our abstraction is not quite right), you could consider refactoring the login process so that it uses an "intermediate" type to model the different authentication phases and only translates the result into a type without an MFA constructor when finished.

msm-code:
Thanks for weighing in! It's 4am here, so I'll address the issues tomorrow.

Do you think that `refactoring the login process so that it uses an "intermediate" type` will make the code cleaner? I'll try to do that tomorrow. I agree that non-exhaustive pattern matches feel hacky, and it's probably a good idea to address them.

jtdaugherty:
As I imagine it, the refactoring would avoid writing the `Setup.hs` case. So, in my opinion, it would result in cleaner code because no call to `error` would be required. I suspect it would lead to other things getting cleaner, too, so I think it's worth exploring.

jtdaugherty:
I'd be okay with holding off on that refactoring until after these changes are merged, too, since the refactoring would be exploratory.

msm-code:
Life happened yesterday, so couldn't work on this until today. 

I've added a new type:
```haskell

-- | The result of a successfull login attempt.
data LoginData =
    LoginSuccess ConnectionData Session User (Maybe Text) --team
    -- ^ Data associated with the new logged-in session.
```

I think it made the code clearer overall (no duplicated case in Login.hs, and no unhandled cases). But naming things is hard, so I'd appreciate your input.

> Thank you for the update! I left comments on the commits requesting a few minor changes. Overall, though, the changes look good to me and I appreciate you putting in more effort on this and updating it to gather the OTP token modally!

I, uhh, I can't see them. Are you sure you clicked the "submit review" button? Happened to me more than once on Github. Otherwise can you guide me to where the comments are? I might be figuratively blind.

jtdaugherty:
It looks like you found the comments and I posted a reply on a couple of them.

msm-code:
You're right, found the comments. Sorry for confusion, for some reason I've assumed PR comments even though you've said `commits`.

Adressde almost all of your comments, except I didn't make ciOTPToken `Maybe Text` yet. To be more precise, I'm struggling with a following type error:

```
src/Matterhorn/Login.hs:452:54: error:
    • Couldn't match type ‘Text’ with ‘Maybe Text’
      Expected type: (Text -> f Text)
                     -> ConnectionInfo -> f ConnectionInfo
        Actual type: (Maybe Text -> f (Maybe Text))
                     -> ConnectionInfo -> f ConnectionInfo
    • In the first argument of ‘editTextField’, namely ‘ciOTPToken’
      In the second argument of ‘(@@=)’, namely
        ‘editTextField ciOTPToken OTPToken (Just 1)’
      In the expression:
        label "OTP Token:" @@= editTextField ciOTPToken OTPToken (Just 1)
    |
452 |     in newForm [label "OTP Token:" @@= editTextField ciOTPToken OTPToken (Just 1)]
```

I understand the problem, just looking for an idiomatic way to fix it. I think I need to find a right lens for the job, but a few first attempts didn't work - I'll get back to this tomorrow.

jtdaugherty:
The issue there, essentially, is that the form field being used expects to deal with a `Text` but the field type in the record is (now) `Maybe Text`. Using a different lens won't help because that would require making an assumption that there is always a value (e.g. with `_Just`) which would crash the program if `Nothing` is in the record field. I think the thing that needs to be done is to adapt the form field to treat `""` as `Nothing` and anything else as `Just`. I'm happy to help with that if you like, but if that is more trouble than you want to deal with, I'd also be okay with skipping the change to `Maybe` in this PR.

jtdaugherty:
I had some time and hacked up an optional field constructor:
```haskell
editOptionalPasswordField :: (Show n, Ord n) => Lens' s (Maybe T.Text) -> n -> s -> FormFieldState s e n
editOptionalPasswordField stLens n =
    let ini Nothing = ""
        ini (Just t) = t
        val ls =
            let stripped = T.strip $ T.concat ls
            in if T.null stripped
               then Just Nothing
               else Just $ Just stripped
        renderVal s = txt $ T.replicate (T.length $ T.concat s) "*"
    in editField stLens n (Just 1) ini val renderVal id
```
You can use that instead of `editPasswordField` when building the form. (This should work just by replacing `editPasswordField` with `editOptionalPasswordField` in the existing code.)

msm-code:
Oh, sorry, I attempted to solve this on my own before noticing your comment. My approach was a bit different - I've created a `wrapMaybe` function to "wrap" ciOTPToken with some default values:

```haskell
mkOTPForm =
    let (...)
        wrapMaybe func = fmap Just . func . (fromMaybe "")
    in newForm [label "OTP Token:" @@= editTextField (ciOTPToken . wrapMaybe) OTPToken (Just 1)]
```

Should've checked Github notifications first. Your code snippet looks more generic and easier to understand, so I'll apply it in a second (with a small change - this is not a password field, and OTP tokens are not usually masked, so I think the name should be `editOptionalTextField`).

jtdaugherty:
Okay, sounds good.

msm-code:
And done.

For reference, this is how it looks like on the web client (notice that the token is not masked with `*`s):

image

jtdaugherty:
Okay, thanks! Let me know when you are ready for this to be merged (i.e. all outstanding changes are finished).

msm-code:
On my side everything is ready, and I've tested a few obvious use-cases like:

- user enter a valid password and a valid token
- user enters invalid password, then a valid password, then invalid token, then a valid token
- user enters a valid password, then exists
- user enters invalid password, then exists
- user exits immediately

Everything seems to work.

One caveat is that there is currently another PR to the mattermost-api submodule from me. It is required for this PR to work. PR link is https://github.com/matterhorn-chat/mattermost-api/pull/56. It was rebased on mater recently.

I've added a commit that bumps the submodule version in this project to the new version (unmerged yet).

I hope that makes sense

msm-code:
Actually I've made a bit of a mess in the mattermost-api commits. Give me a second and I'll squash the commits.

msm-code:
Done. If you prefer, I can squash commits in this PR too.

jtdaugherty:
No thanks on the squash. I don’t want to lose history unless the intermediate changes are broken or exploratory. Thank you for your work on this!

PR #664, opened 2020-09-10
Fix broken relative urls in UserGuide.md

nazywam:
Hey, while browsing https://github.com/matterhorn-chat/matterhorn/blob/master/docs/UserGuide.md I've noticed that those two relative links to other doc files seem to be broken.


jtdaugherty:
Thank you!

PR #689, opened 2021-02-15
favorite channel feature

ajayeeralla:


jtdaugherty:
Just to record this here explicitly: I checked offline, and @ajayeeralla is not yet ready for this to be reviewed.

jtdaugherty:
Hi @ajayeeralla, how is this work going? Is there anything we can do to help?

ajayeeralla:
> Hi @ajayeeralla, how is this work going? Is there anything we can do to help?

Hi @jtdaugherty, I have not made any progress on this yet. Soon, I'll let you know, may be in a couple of days. Thanks.

jtdaugherty:
Okay, thanks - no hurry!

src/Matterhorn/State/Channels.hs
```diff
@@ -1083,3 +1084,26 @@ updateChannelNotifyProps cId notifyProps = do
             Just tId -> mh $ invalidateCacheEntry $ ChannelSidebar tId
 
         csChannel(cId).ccInfo.cdNotifyProps .= notifyProps
+
+toggleChannelFavoriteStatus :: MH()
```
jtdaugherty, line 1088: This is a style note: the type signature here should be `MH ()`, with a space between `MH` and `()`.

src/Matterhorn/State/Channels.hs
```diff
@@ -1083,3 +1084,26 @@ updateChannelNotifyProps cId notifyProps = do
...
+        trueVal = "true"
+        prefVal =  case favPref of
+            Just True -> ""
+            Just False -> trueVal
+            Nothing -> trueVal
+        pref = Preference 
+            { preferenceUserId = myId
+            , preferenceCategory = PreferenceCategoryFavoriteChannelShow
+            , preferenceName = PreferenceName $ idString cId
+            , preferenceValue = PreferenceValue $ prefVal
```
jtdaugherty, line 1105: The `$` is not needed here any longer.

src/Matterhorn/Types.hs
```diff
@@ -251,6 +251,7 @@ module Matterhorn.Types
   , userPrefTeamOrder
   , dmChannelShowPreference
   , groupChannelShowPreference
+  , favoriteChannelPreference
```
jtdaugherty, line 254: Thank you for renaming this! In the future, please make changes like this in separate commits. The rename of this function is not related to the addition of `toggleChannelFavoriteStatus`. Having these changes in the same commit usually means they are related in some way, but as far as I can tell they could be committed separately.

src/Matterhorn/State/Channels.hs
```diff
@@ -588,6 +588,25 @@ applyPreferenceChange pref = do
                       Nothing -> return ()
               False -> do
                   csChannel(cId).ccInfo.cdSidebarShowOverride .= Nothing
+      
+      | Just f <- preferenceToFavoriteChannelPreference pref -> do
```
jtdaugherty, line 593: This case arm is not needed. This logic is important for DM and group channel preference changes when we get updates from the server to show those channels in the sidebar. We have this logic so that we can tell when those changes were triggered by the Matterhorn user in an attempt to switch to those channels. But for favorite status changes, we don't need this logic. I'm happy to talk about it in more detail if you like. But this hunk should be removed.

src/Matterhorn/Command.hs
```diff
@@ -277,12 +277,20 @@ commandList =
   , Cmd "move-team-right" "Move the currently-selected team to the right in the team list" NoArg $ \_ ->
         moveCurrentTeamRight
   
-  --, Cmd "favorite-channel" "Mark the current channel as favorite/non-favorite" NoArg $ \_ -> do
-  --      tId <- use csCurrentTeamId
-  --      cId <- use (csCurrentChannelId tId)
-  --      s <- getSession
-  --    pref <- use crUserPreferences
-  --   applyPreferenceChange pref
+  , Cmd "favorite-channel" "Mark the current channel as favorite/non-favorite" NoArg $ \_ -> do
```
jtdaugherty: Since this command implementation was temporary as you got things working, I recommend not committing changes like this in the future. The reasoning is that these changes are helpful during development, but they make patch review difficult since the reviewer needs to follow along to see whether these changes were replaced with other changes later in the patch sequence. I suggest using `git add -p` and skipping these hunks when committing. In general, I recommend only committing changes that are working, and keeping them as uncommitted until they have been finished. Let me know if it would be helpful to discuss that more; this is very important!

src/Matterhorn/Command.hs
```diff
@@ -276,6 +276,13 @@ commandList =
 
   , Cmd "move-team-right" "Move the currently-selected team to the right in the team list" NoArg $ \_ ->
         moveCurrentTeamRight
+  
+  --, Cmd "favorite-channel" "Mark the current channel as favorite/non-favorite" NoArg $ \_ -> do
```
jtdaugherty: Similarly to my comment on this part of the code in a different patch, please leave comment hunks like this out of your commits. I know it can be helpful to keep those around for your own development, but they don't need to be committed (and can lead to confusion).

jtdaugherty:
I have another overall comment about the patches, which is that it would be very helpful if the commit messages can specifically describe the changes in each commit. In cases where there is a clear reason for the change, it would also be helpful if that reason was included. For many of the commits in this set of patches, the description does not match the contents and doesn't provide a description of why the changes were made. I have a resource on how to do this that I'll send you directly.

src/Matterhorn/State/Channels.hs
ajayeeralla: Done.

src/Matterhorn/State/Channels.hs
ajayeeralla: Noted.

src/Matterhorn/Types.hs
ajayeeralla: Noted.

src/Matterhorn/State/Channels.hs
ajayeeralla: @jtdaugherty if we remove this arm I checked that toggling favorite channel status doesn't work as we desired. I would like to discuss tomorrow if you are available

ajayeeralla: @jtdaugherty if we remove this arm I checked that toggling favorite channel status doesn't work as we desired. I would like to discuss tomorrow if you are available

src/Matterhorn/Command.hs
ajayeeralla: Noted.

src/Matterhorn/Command.hs
ajayeeralla: Noted.

jtdaugherty:
@ajayeeralla I found a bug in this branch:
1. Go to a DM channel
2. Run `/favorite-channel`
3. The channel disappears from the list

It looks like the issue is in `Types.getFavoriteChannelEntriesInOrder`: it checks for `info^.ccInfo.cdTeamId == Just tId`, but DM channels have a team ID of `Nothing`. So the check should be `(info^.ccInfo.cdTeamId == Just tId || info^.ccInfo.cdTeamId == Nothing) && ...`

src/Matterhorn/State/Channels.hs
jtdaugherty: I looked into this, and the issue is that we do need the call to `updateSidebar`, but we don't need the rest of the code in this arm. We need the `updateSidebar` call because that forces a redraw of the channel sidebar whenever a channel's favorite status changes. The favorite feature still works without that, but we just don't see the sidebar update.

src/Matterhorn/State/Channels.hs
```diff
@@ -1100,9 +1100,9 @@ toggleChannelFavoriteStatus = do
             Nothing -> trueVal
         pref = Preference 
             { preferenceUserId = myId
-            , preferenceCategory = PreferenceCategoryFavoriteChannelShow
+            , preferenceCategory = PreferenceCategoryFavoriteChannel
```
jtdaugherty, line 1103: This is an example of a change that belongs in another commit (you had a commit that did a lot of the other renamings of this constructor). How do you stage your commits? If you use your editor, does it have support for staging specific hunks? If you are using `git` on the command line, are you familiar with `git add -p`?

Issue #628, opened 2020-06-25
Show attachment file browser keybindings in keybinding help


jtdaugherty:
Right now the keybindings available in the attachment disk browser are not included in the keybinding help and so they are not discoverable.

jtdaugherty:
Here's a summary of some next steps:

* Factor out and export functions corresponding to key-specific event handler's in `brick`'s `FileBrowser.hs` module (for "normal" and "common" events; the "searching" events should probably stay opaque)
* Add `KeyEvent` constructors for each new event that we need to handle, and add default keybindings for each to `Keybindings.hs`
* Add keyboard handlers for the abstract events to `attachmentBrowseKeyHandlers` in `Matterhorn.Events.ManageAttachments`

PR #690, opened 2021-02-16
Feature/expose file browser key bindings

TieDyedDevil:
Resolve issue #628.


jtdaugherty:
I'm going to close this in favor of @TieDyedDevil's other branch, `feature/file-browser-key-bindings-628`, which is based on `develop` (not `master`, as is the case here).

PR #696, opened 2021-03-04
Add /attach command

karljs:
This adds an /attach command, which allows for attaching a file to a message by giving a path directly instead of navigating the file browser.  In order to share code, a few small things moved from the files in the Events directory to ones in the State directory.  As of the first commit, it does not work and is only here to ease discussion.

### DIFF
--- a/src/Matterhorn/Events.hs
+++ b/src/Matterhorn/Events.hs
@@ -157,9 +157,9 @@ formatError (NoSuchHelpTopic topic) =
     let knownTopics = ("  - " <>) <$> helpTopicName <$> helpTopics
     in "Unknown help topic: `" <> topic <> "`. " <>
        (T.unlines $ "Available topics are:" : knownTopics)
-formatError (BadAttachmentPath e) =
+formatError (AttachmentException e) =
     case fromException e of
-      Just (ioe :: IO.IOError) -> 
+      Just (ioe :: IO.IOError) ->
           if IO.isDoesNotExistError ioe
           then "Error attaching, file does not exist!"
           else if IO.isPermissionError ioe
@@ -168,6 +168,8 @@ formatError (BadAttachmentPath e) =
       Nothing -> "Unknown error attaching file!\n" <>
           "Please report this error at https://github.com/matterhorn-chat/matterhorn/issues"
           -- this case shouldn't be reached
+formatError (BadAttachmentPath msg) = 
+    msg
 formatError (AsyncErrEvent e) =
     "An unexpected error has occurred! The exception encountered was:\n  " <>
     T.pack (show e) <>
--- a/src/Matterhorn/Events.hs
+++ b/src/Matterhorn/Events.hs
@@ -155,6 +155,8 @@ formatError (NoSuchHelpTopic topic) =
     let knownTopics = ("  - " <>) <$> helpTopicName <$> helpTopics
     in "Unknown help topic: `" <> topic <> "`. " <>
        (T.unlines $ "Available topics are:" : knownTopics)
+formatError (BadFile file) =
+    "Specified file " <> file <> " either doesn't exist or couldn't be read"
 formatError (AsyncErrEvent e) =
     "An unexpected error has occurred! The exception encountered was:\n  " <>
     T.pack (show e) <>
--- a/src/Matterhorn/Events.hs
+++ b/src/Matterhorn/Events.hs
@@ -10,9 +10,10 @@ import           Matterhorn.Prelude
 
 import           Brick
 import qualified Data.Text as T
-import           GHC.Exception.Type ( displayException )
+import           GHC.Exception.Type ( fromException )
 import qualified Graphics.Vty as Vty
 import           Lens.Micro.Platform ( (.=), _2, singular, _Just )
+import qualified System.IO.Error as IO
 
 import qualified Network.Mattermost.Endpoints as MM
 import           Network.Mattermost.Exceptions ( mattermostErrorMessage )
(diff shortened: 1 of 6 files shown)


src/Matterhorn/Types.hs
```diff
@@ -1917,6 +1917,9 @@ data MHError =
     -- ^ The specified script was not found
     | NoSuchHelpTopic T.Text
     -- ^ The specified help topic was not found
+    | BadFile T.Text
```
jtdaugherty, line 1920: A suggestion on this: rename to `BadAttachmentPath` to be more specific.

jtdaugherty: (and maybe consider including the exception, or more detail about the nature of the badness)

src/Matterhorn/State/Attachments.hs
```diff
@@ -54,3 +62,50 @@ showAttachmentFileBrowser = do
...
+        case Vector.find matches es of
+            Just _ -> return ()
+            Nothing -> do
+                tryReadAttachment entry >>= \case
+                    Right a -> do
+                        oldIdx <- use (csCurrentTeam.tsEditState.cedAttachmentList.L.listSelectedL)
+                        let newIdx = if Vector.null es
+                                     then Just 0
+                                     else oldIdx
+                        csCurrentTeam.tsEditState.cedAttachmentList %= L.listReplace (Vector.snoc es a) newIdx
```
jtdaugherty, line 93: This right here is the bit that needs to be run whenever an attachment is added. `attachFileByPath` doesn't do this, and it's the key part of the state change that makes it so the attachment gets sent to the server. The `cedAttachmentList` is what is used to build the message when it is sent.

jtdaugherty: (by "build the message" I mean "include the attachments when building the message")

jtdaugherty: To see the usage of this, see `Matterhorn.State.handleInputSubmission`.

src/Matterhorn/Types.hs
jtdaugherty: (and then the code to construct the error message text can just build the right text for the error)

karljs: Thanks, made those improvements!

src/Matterhorn/Events.hs
```diff
@@ -155,8 +156,8 @@ formatError (NoSuchHelpTopic topic) =
     let knownTopics = ("  - " <>) <$> helpTopicName <$> helpTopics
     in "Unknown help topic: `" <> topic <> "`. " <>
        (T.unlines $ "Available topics are:" : knownTopics)
-formatError (BadFile file) =
-    "Specified file " <> file <> " either doesn't exist or couldn't be read"
+formatError (BadAttachmentPath e msg) =
+    msg <> " The exception encountered was:\n " <> T.pack (displayException e)
```
jtdaugherty, line 160: Would it be possible to make this message inspect the exception? These functions come to mind:

https://hackage.haskell.org/package/base-4.14.1.0/docs/System-IO-Error.html#g:2

I'm thinking that the raw `Show` of the exception will probably not be what the user wants to see (and may not know how to act on it), but if we can translate common exceptions into more user-friendly language, that would be nice. (An easy one is "file not found")

src/Matterhorn/State/Attachments.hs
```diff
@@ -68,9 +68,8 @@ attachFileByPath txtPath = do
     let strPath = unpack txtPath
     fileInfo <- liftIO $ FB.getFileInfo strPath strPath
     case FB.fileInfoFileStatus fileInfo of
-        Left _ -> do
-            mhLog LogError $ "Failed to obtain FileInfo for file: " <> txtPath
-            mhError $ BadFile txtPath
+        Left e -> do
+            mhError $ BadAttachmentPath (toException e) "Unable to stat the requested file.  Check that it exists and has proper permissions"
```
jtdaugherty, line 72: I'm concerned that the word "stat" here won't be well-known to users. Could we say something similar, like "Unable to attach the requested file. Check that it exists and has proper permissions."?

src/Matterhorn/State/Attachments.hs
karljs: I'm not sure I understand completely.  It looks to me like this code does get called, provided no errors arise.  `attachFileByPath` ultimately delegates the work to `tryAddAttachment` (which is also the function called by the attachment browsing code).

Locally, I used a bunch of logging commands to debug this.  I tried outputting the length of the attachment list vector before and after the command is executed, and it seemed to be working (moving from 0 to 1).

jtdaugherty: Oh, I was mistaken - I didn't see the call to `tryAddAttachment` in `attachFileByPath`. Hmm.

src/Matterhorn/State/Attachments.hs
karljs: This should be addressed now

src/Matterhorn/Events.hs
karljs: I've done a small refactor to store only the exception instead of the message.  Now the messages are generated by querying the exception.

src/Matterhorn/Events.hs
```diff
@@ -156,8 +157,17 @@ formatError (NoSuchHelpTopic topic) =
...
+formatError (BadAttachmentPath e) =
+    case fromException e of
+      Just (ioe :: IO.IOError) -> 
+          if IO.isDoesNotExistError ioe
+          then "Error attaching, file does not exist!"
+          else if IO.isPermissionError ioe
+               then "Error attaching, lacking permissions to read file!"
+               else "Unable to attach the requested file.  Check that it exists and has proper permissions."
+      Nothing -> "Unknown error attaching file!\n" <>
+          "Please report this error at https://github.com/matterhorn-chat/matterhorn/issues"
```
jtdaugherty, line 169: In this case, could you `mhLog` the value of `show e` so users can give us that when reporting?

karljs: `formatError` is a pure function.  I can do this but it will necessarily involve a somewhat larger change.  Is that okay?

jtdaugherty: To confirm here: no need to refactor this, please disregard my `mhLog` suggestion.

karljs:
Removing WIP status as this now appears to work as intended

PR #699, opened 2021-03-12
Update user metadata correctly

### LINKED ISSUE #697: Consistently update user metadata when it changes
I just noticed that a colleague still had a `[OOO]` tag in their nickname to show he was out of office. After notifying him, he mentioned it was already removed. After quitting and reopening matterhorn the new nickname was picked up indeed.

Is it possible to have it update the nickname every now and then? I'm usually running matterhorn under tmux, so it's running for a few days at a time.

karljs:
This implements a working handler for user metadata.  Previously we were ignoring those websocket events.  Now we correctly update the displayed user information (e.g., nickname) as soon as this happens.  This comes with a related change to `mattermost-api` that changes some of the logging-related types, since those were part of figuring out this fix.

Closes #697

### DIFF
--- a/src/Matterhorn/Events/Websocket.hs
+++ b/src/Matterhorn/Events/Websocket.hs
@@ -11,7 +11,7 @@ import qualified Data.HashMap.Strict as HM
 import qualified Data.Sequence as Seq
 import qualified Data.Set as Set
 import qualified Data.Text as T
-import           Lens.Micro.Platform ( (%=), preuse )
+import           Lens.Micro.Platform ( preuse )
 
 import           Network.Mattermost.Lenses
 import           Network.Mattermost.Types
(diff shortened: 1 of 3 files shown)


src/Matterhorn/Events/Websocket.hs
```diff
@@ -214,7 +214,13 @@ handleWebsocketEvent we = do
             mhLog LogGeneral $ T.pack $
                 "WMTeamDeleted event: " <> show we
 
-        WMUserUpdated -> return ()
+        WMUserUpdated
+            | Just user <- wepUser (weData we) -> do
```
jtdaugherty, line 218: Could you factor the body of this case arm out to a function that lives in `Matterhorn.State.Users`? That way this can be abstract w.r.t. the details of the update.

karljs: I did what I could here, but the `Matterhorn.State.Users` module can't refresh the channel without bigger changes due to cyclic dependencies.  I think just leaving the refresh call with the event is okay, but if you have a better suggestion I'm happy to adjust that.

jtdaugherty: Okay, that makes sense. We already have a handful of `boot` files for resolving import cycles so I don't think it would be much work to fully factor this code out, but I'm okay with it as it is in this patch set. Thanks!

PR #701, opened 2021-03-17
trim channel autocomplete topics to one line

karljs:


src/Matterhorn/Draw/Autocomplete.hs
```diff
@@ -190,7 +191,7 @@ renderChannelCompletion c inChan selected =
             , padTo displayNameWidth $
               withDefAttr channelNameAttr $
               txt $ sanitizeUserText $ channelDisplayName c
-            , txt $ sanitizeUserText $ channelPurpose c
+            , txt $ trimToOneLine $ sanitizeUserText $ channelPurpose c
```
jtdaugherty, line 194: I suggest `vLimit 1 $ txt $ ...` here, which limits the brick rendering to 1 row high, rather than taking the first physical line of text. That captures the intention while decoupling it from the text representation; that would be helpful if the text rendering were to change (such that even a single physical line of text takes up multiple rows of screen space).

karljs: Good call, done!

PR #712, opened 2021-05-17
Feature/json notifier

### LINKED ISSUE #681: Provide additional information to notification commands
Right now the notification script API entails providing three command-line arguments to the notification command:
* A value of `1` to indicate that the user was mentioned, `2` otherwise
* The username of the message sender
* The message body

In addition there are other bits of information that it would be useful to provide, such as:
* A flag indicating whether the message was a DM
* A channel name, if possible
* The name of the team associated with the channel, if any (DMs have no associated team)

TieDyedDevil:
Resolves #681.

### DIFF
--- a/docs/notification-scripts.md
+++ b/docs/notification-scripts.md
@@ -17,7 +17,7 @@ The interface for notification scripts is as follows:
 * The notifier behavior changes according to the activityNotifyVersion
   setting in the configuration file.
 
-* Version 1: Matterhorn will invoke the command with three arguments:
+* NotifyV1: Matterhorn will invoke the command with three arguments:
   * The mention argument, whose values are:
     * "1" - this value indicates that the user running Matterhorn was
       mentioned in the message body.
@@ -28,9 +28,9 @@ The interface for notification scripts is as follows:
   * The message body, sanitized of tabs (converted to spaces) and escape
     characters (converted to "<ESC>").
 
-* Version 2: Matterhorn will invoke the command with zero arguments,
+* NotifyV2: Matterhorn will invoke the command with zero arguments,
   passing notification details via JSON to the command's standard input.
-  * Version 2 JSON fields:
+  * NotifyV2 JSON payload fields:
     * "version": "2"
     * "from": "sender name"
     * "message": "message text..."
@@ -38,7 +38,7 @@ The interface for notification scripts is as follows:
 
 * Matterhorn will wait for the process to terminate. If the process
   emits any output to standard out OR if the command exits with a
-  non-zero xit status, Matterhorn will consider that evidence that
+  non-zero exit status, Matterhorn will consider that evidence that
   the command has failed and will display an error to the user in the
   current channel. The standard out and standard error output will be
   logged to a temporary subprocess log file and the log file's path will
--- a/docs/notification-scripts.md
+++ b/docs/notification-scripts.md
@@ -14,7 +14,10 @@ The interface for notification scripts is as follows:
   to finish, which will cause Matterhorn to delay looking at incoming
   network messages, etc.
 
-* Matterhorn will invoke the command with three arguments:
+* The notifier behavior changes according to the notifier version
+  specified in the configuration file.
+
+* Version 1: Matterhorn will invoke the command with three arguments:
   * The mention argument, whose values are:
     * "1" - this value indicates that the user running Matterhorn was
       mentioned in the message body.
@@ -24,6 +27,10 @@ The interface for notification scripts is as follows:
     message.
   * The message body, sanitized of tabs (converted to spaces) and escape
     characters (converted to "<ESC>").
+
+* Version 2 and later: Matterhorn will invoke the command with zero
+  arguments and pass notification data via JSON to the command's
+  standard input.
 
 * Matterhorn will wait for the process to terminate. If the process
   emits any output to standard out OR if the command exits with a
--- a/docs/notification-scripts.md
+++ b/docs/notification-scripts.md
@@ -28,9 +28,8 @@ The interface for notification scripts is as follows:
   * The message body, sanitized of tabs (converted to spaces) and escape
     characters (converted to "<ESC>").
 
-* Version 2 and later: Matterhorn will invoke the command with zero
-  arguments and pass notification data via JSON to the command's
-  standard input.
+* Version 2: Matterhorn will invoke the command with zero arguments,
+  passing notification details via JSON to the command's standard input.
   * Version 2 JSON fields:
     * "version": "2"
     * "sender": "sender name"
--- a/docs/notification-scripts.md
+++ b/docs/notification-scripts.md
@@ -32,9 +32,9 @@ The interface for notification scripts is as follows:
   passing notification details via JSON to the command's standard input.
   * Version 2 JSON fields:
     * "version": "2"
-    * "sender": "sender name"
+    * "from": "sender name"
     * "message": "message text..."
-    * "mentioned": boolean (true or false)
+    * "mention": boolean (true or false: did sender mention me?)
 
 * Matterhorn will wait for the process to terminate. If the process
   emits any output to standard out OR if the command exits with a
--- a/docs/notification-scripts.md
+++ b/docs/notification-scripts.md
@@ -14,8 +14,8 @@ The interface for notification scripts is as follows:
   to finish, which will cause Matterhorn to delay looking at incoming
   network messages, etc.
 
-* The notifier behavior changes according to the notifier version
-  specified in the configuration file.
+* The notifier behavior changes according to the activityNotifyVersion
+  setting in the configuration file.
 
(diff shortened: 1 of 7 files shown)


docs/sample-config.ini
```diff
@@ -163,6 +163,11 @@
 #
 # activityNotifyCommand = /path/to/notify
 
+# Notification script control: select the notifier payload version.
+# This affects how the activityNotifyCommand is invoked.
```
jtdaugherty, line 167: Can this documentation be updated to list the currently-acceptable values?

docs/notification-scripts.md
```diff
@@ -14,7 +14,10 @@ The interface for notification scripts is as follows:
   to finish, which will cause Matterhorn to delay looking at incoming
   network messages, etc.
 
-* Matterhorn will invoke the command with three arguments:
+* The notifier behavior changes according to the notifier version
+  specified in the configuration file.
```
jtdaugherty, line 18: Can this mention the new `activityNotifyVersion` setting explicitly?

docs/notification-scripts.md
```diff
@@ -25,6 +28,15 @@ The interface for notification scripts is as follows:
   * The message body, sanitized of tabs (converted to spaces) and escape
     characters (converted to "<ESC>").
 
+* Version 2 and later: Matterhorn will invoke the command with zero
```
jtdaugherty, line 31: This should probably say "Version 2" (not "and later") since "2" is the only other supported version at this point.

notification-scripts/notifyV2
```diff
@@ -0,0 +1,56 @@
+#! /usr/bin/env sh
```
jtdaugherty, line 1: How will this script behave in the absence of `jq`? Will it fail gracefully?

src/Matterhorn/State/Messages.hs
```diff
@@ -681,6 +686,57 @@ runNotifyCommand post mentioned = do
                                  [notified, sender, messageString] Nothing Nothing
                 return Nothing
 
+encodeToJSONstring :: A.ToJSON a => a -> String
+encodeToJSONstring a = map (chr . fromEnum) . BL.unpack $ A.encode a
```
jtdaugherty, line 690: Rather than `map (...) . BL.unpack`, please use [Data.ByteString.Char8.unpack](https://hackage.haskell.org/package/bytestring-0.11.1.0/docs/Data-ByteString-Char8.html#v:unpack) to go directly to `String`.

src/Matterhorn/State/Messages.hs
```diff
@@ -681,6 +686,57 @@ runNotifyCommand post mentioned = do
...
+
+-- We define a notifyGetPayload for each notification version.
+notifyGetPayload :: Int -> ChatState -> Post -> Bool -> MH (Maybe String)
```
jtdaugherty, line 693: Style note: this type signature should go on line 710 (immediately preceding the definition).

src/Matterhorn/State/Messages.hs
```diff
@@ -681,6 +686,57 @@ runNotifyCommand post mentioned = do
...
+
+-- Notification Version 2
+data NotificationV2 = NotificationV2
+    { version :: Int
+    , message :: Text
+    , mentioned :: String
```
jtdaugherty, line 699: If I recall, this is a boolean value. If that's right, then I think this field would be better as a `Bool` with a conversion to the `yes`/`no` values moved to the en/decoder.

jtdaugherty: In fact, really, it would be okay to just put a bool in the JSON data rather than encoding as a string.

src/Matterhorn/State/Messages.hs
```diff
@@ -681,6 +686,57 @@ runNotifyCommand post mentioned = do
...
+    , sender :: Text
+    } deriving (Show)
+
+instance A.ToJSON NotificationV2 where
+    toJSON ( NotificationV2 version message mentioned sender ) =
```
jtdaugherty, line 704: Style note: no whitespace around the pattern-matched constructor here. (So just `(NotificationV2 ... sender)`).

src/Matterhorn/State/Messages.hs
```diff
@@ -681,6 +686,57 @@ runNotifyCommand post mentioned = do
...
+        A.object [ "version"    A..= version
+                 , "message"    A..= message
+                 , "mentioned"  A..= mentioned
+                 , "sender"     A..= sender
+                 ]
+
+notifyGetPayload 2 st post mentioned = do
```
jtdaugherty, line 711: There are a bunch of compiler warnings in this model that I'm wondering if you could address. This one I spotted myself: this function will cause a runtime exception if invoked on any payload version argument other than `2` because there is no fall-through case.

Also, a separate design thought: if the valid versions are only 1 and 2, then I think it would be easier to avoid issues with handling all possible version values exhaustively by making a new `NotificationVersion` type, e.g. `data NotificationVersion = NotifyV1 | NotifyV2`. This has the benefit that pattern-matches on this type can be checked for exhaustiveness, which is a very desirable thing to do whenever possible.

jtdaugherty: (It also prevents anyone from mistaking this `Int` to be useful in a spot where some other non-version `Int` value is used.)

src/Matterhorn/State/Messages.hs
jtdaugherty: Another thing I noticed about this function is that it is used in the `MH` monad, but that's unnecessary at present since all it does is `return`. It could be rewritten as a pure function (having type `... -> Maybe String`) and then used in a `let` binding below rather than with a monadic bind (`<-`).

docs/sample-config.ini
TieDyedDevil: Resolved by ba908325.

docs/notification-scripts.md
TieDyedDevil: Resolved by commit 111ba526 .

docs/notification-scripts.md
TieDyedDevil: Resolved by commit 07fb71e2 .

notification-scripts/notifyV2
TieDyedDevil: Resolved by commit 60f4c041 .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit f96534a4 .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit 2b0420c0 .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit 6553fb5b .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit 54d3635c .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit 69f9a387 .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commits f96534a and adc90b78 .

docs/notification-scripts.md
TieDyedDevil: Resolved by commit 111ba52 .

docs/notification-scripts.md
TieDyedDevil: Resolved by commit 07fb71e .

notification-scripts/notifyV2
TieDyedDevil: Resolved by commit 60f4c04 .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit f96534a .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit 6553fb5 .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit 54d3635 .

src/Matterhorn/State/Messages.hs
TieDyedDevil: Resolved by commit 69f9a38 .

jtdaugherty:
Thank you @TieDyedDevil!

PR #722, opened 2021-06-29
Relax some upper version bounds to allow building with GHC 9.0

RyanGlScott:
This relaxes the upper version bounds for `aeson` and `strict` to allow `matterhorn` to build against versions `1.5.*` and `0.4.*`, respectively, which are the minimum versions that build against GHC 9.0.


jtdaugherty:
Thank you!

Issue #683, opened 2020-12-19
Using Paths_matterhorn for data files


VannTen:
Would you consider using Paths_matterhorn (the module generated by Cabal) for
`data-files` ? (and adding emoji.json to those)

Currently the data-files are located at the same place than the executable.

# Benefits

That would allow the install / packager to specify where to place these files ;
that would be make it easier to have native packages for various distributions.

# Problems
However, it might cause some problems with the development build and the current
way of doing releases.

I've started to modify the code to see how it would work and it does break at
least ./run.sh, but it works when I use `cabal run`. I understand that cabal
does not set the correct path in the auto generated module, but use env variables
in `run` to resolve to the correct path (`matterhorn_datadir` etc).

For binary releases, I assume it would also break, since during the build there
is now way to know where the tarbal will be extracted.

---------------------------------

An option would be to support both the current mechanism and the
Paths_matterhorn and the current mechanism but that adds some cognitive
overhead.

# Conclusion

Giving the disruption it would cause, I will understand if you do not wish to
merge such a change, and I'll maintain a patch on my own.

P.S. : By the way, Matterhorn works greats :).

karljs:
I've investigated the implications of this suggestion.  It seems to be there are 4 different workflows we would ideally support: 
1. Users downloading the existing binary releases
2. Package maintainers trying to package for specific distros and repositories
3. Folks running development builds, such as via `run.sh`
4. Developers loading code in a `cabal repl`

Using a `Paths_matterhorn` module, in principle, appears to support these as follows:
1. Not at all
2. Fine, evidently.  I don't have personal experience with this.
3. Not well, because if we run the executable without cabal then the paths don't seem to be set correctly. 
4. Fine

Our current approach (look at the path of the executable and determine where to look for data files ourselves) seems to support the following:
1. Fine
2. Not well because it assumes data files will be in a sibling directory to the executable
3. Fine
4. Fine

@jtdaugherty Assuming you agree that we want to support those workflows, I think the thing to do might be to keep the current implementation as a default, but add a configuration flag to the cabal file that enables the use of a `Paths_matterhorn` module as a separate code path.  That way, package maintainers have a way of achieving what they need but we don't disrupt any of the current workflows that folks depend on.

jtdaugherty:
[...]

PR #724, opened 2021-07-13
Check cabal-installed data directory for syntax files and emoji.json

sternenseemann:
When installed using cabal, we can use the `Paths_matterhorn` module and `getDataFileName` to get the paths to files and directories which are listed in `data-files` in `matterhorn.cabal`.

This fixes wrong assumptions of matterhorn when it is installed in a different way than the official release build. For example in NixOS we install data files to a separate location cabal knows about, so this change is necessary to make for example emoji completion work on NixOS: https://github.com/NixOS/nixpkgs/issues/129048.

I'd suggest you switch to `getDataFileName` completely, but since I don't know anything about the release build process, I can't implement such a change.

### DIFF
--- a/matterhorn.cabal
+++ b/matterhorn.cabal
@@ -17,6 +17,9 @@ tested-with:         GHC == 8.0.2, GHC == 8.2.2, GHC == 8.4.4, GHC == 8.6.5, GHC
 data-files:          syntax/*.xml
                      syntax/language.dtd
                      syntax/LICENSE
+                     emoji/NOTICE.txt
+                     emoji/LICENSE.txt
+                     emoji/emoji.json
 
 extra-doc-files:     CHANGELOG.md
                      README.md
--- a/src/Matterhorn/Config.hs
+++ b/src/Matterhorn/Config.hs
@@ -13,6 +13,8 @@ where
 import           Prelude ()
 import           Matterhorn.Prelude
 
+import qualified Paths_matterhorn as Paths
+
 import           Control.Monad.Trans.Except
 import           Control.Monad.Trans.Class ( lift )
 import           Data.Char ( isDigit, isAlpha )
@@ -48,7 +50,8 @@ defaultSkylightingPaths :: IO [FilePath]
 defaultSkylightingPaths = do
     xdg <- xdgSyntaxDir
     adjacent <- getBundledSyntaxPath
-    return [xdg, adjacent]
+    cabalDataFiles <- Paths.getDataFileName syntaxDirName
+    return [xdg, adjacent, cabalDataFiles]
 
 getBundledSyntaxPath :: IO FilePath
 getBundledSyntaxPath = do
--- a/src/Matterhorn/FilePaths.hs
+++ b/src/Matterhorn/FilePaths.hs
@@ -25,6 +25,8 @@ where
 import Prelude ()
 import Matterhorn.Prelude
 
+import qualified Paths_matterhorn as Paths
+
 import Data.Text ( unpack )
 import System.Directory ( doesFileExist
                         , doesDirectoryExist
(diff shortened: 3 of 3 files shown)


matterhorn.cabal
```diff
@@ -17,6 +17,8 @@ tested-with:         GHC == 8.0.2, GHC == 8.2.2, GHC == 8.4.4, GHC == 8.6.5, GHC
 data-files:          syntax/*.xml
                      syntax/language.dtd
                      syntax/LICENSE
+                     emoji/*.txt
```
jtdaugherty, line 20: Please list all files explicitly here rather than using a wildcard. I don't want this wildcard accidentally including things that get added to that directory that aren't intended to be installed.

src/Matterhorn/FilePaths.hs
```diff
@@ -87,7 +94,9 @@ bundledEmojiJsonPath = do
                  -- In this case we assume the binary is being run from
                  -- a release, in which case the syntax directory is a
                  -- sibling of the executable path.
-                 takeDirectory selfPath </> emojiDirName </> emojiJsonFilename
+                 fromMaybe
+                     (takeDirectory selfPath </> emojiDirName </> emojiJsonFilename)
+                     dataEmojiJsonPath
```
jtdaugherty, line 99: Please flip the order on this `fromMaybe` so that the preferred path is the previously-used one and the fallback path is `dataEmojiJsonPath`.

src/Matterhorn/Config.hs
```diff
@@ -47,8 +49,9 @@ userSyntaxPlaceholderName = "USER_SYNTAX"
 defaultSkylightingPaths :: IO [FilePath]
 defaultSkylightingPaths = do
     xdg <- xdgSyntaxDir
+    cabalDataFiles <- Paths.getDataFileName syntaxDirName
     adjacent <- getBundledSyntaxPath
-    return [xdg, adjacent]
+    return [xdg, cabalDataFiles, adjacent]
```
jtdaugherty, line 54: Please put `cabalDataFiles` last in this list.

jtdaugherty:
Thanks for your patch! I don't know if you consulted https://github.com/matterhorn-chat/matterhorn/issues/683 at all while doing this, but if you didn't, then this change happens to coincide with the discussion there. (In the future, please consider opening an issue or commenting on one first before writing a patch to increase the chances that we'll merge it; patches like this without discussion beforehand can be more challenging to merge if we don't agree with the decisions made in the patch.)

Once you've had a chance to address my comments, I'll merge. Thanks!

matterhorn.cabal
sternenseemann: Done.

src/Matterhorn/FilePaths.hs
sternenseemann: Fair enough, done.

src/Matterhorn/Config.hs
sternenseemann: Done.

sternenseemann:
> Thanks for your patch! I don't know if you consulted #683 at all while doing this, but if you didn't, then this change happens to coincide with the discussion there. (In the future, please consider opening an issue or commenting on one first before writing a patch to increase the chances that we'll merge it; patches like this without discussion beforehand can be more challenging to merge if we don't agree with the decisions made in the patch.)

In my case I was trying to fix an issue we had in packaging by upstreaming support for the default cabal data files mechanism, so me having to apply this as a patch would've been an acceptable outcome as well. But I do think that upstreaming patches is beneficial for everyone involved usually :)

jtdaugherty:
That makes sense to me, @sternenseemann - I'd definitely like to find a way to get changes merged rather than have others have to maintain their own patches or forks. Thanks!

PR #789, opened 2022-12-29
Make it possible to get OTP tokens using otptokencmd

### LINKED ISSUE #753: Support for `otpcmd` in the config.ini
Follow up for #228, especially [this comment](https://github.com/matterhorn-chat/matterhorn/issues/228#issuecomment-997544747).

As I've said, I personally don't think I'll use this feature, but I'm happy to implement it since it's closely related to my [previous PR](https://github.com/matterhorn-chat/matterhorn/pull/663).

1. Can one of the maintainers (@jtdaugherty I suppose) confirm that it's OK to implement? if yes, I'd like to work on it.
2. Do I understand correctly, that the behaviour should mirror that of `passcmd` and `tokencmd`, but the output of the command should be used as the 2fa token? The configuration key @jtdaugherty suggested is `otptokencmd`
3. is there anything special I should keep in mind? Especially @toastal and @eyenx.

msm-code:
Sorry for letting this wait for so long - I had an almost ready PR long time ago and somehow didn't share it.

Works for me with the following configuration:

```
otptokencmd: cloak view my_account_name
```

(I use https://github.com/evansmurithi/cloak for OTP codes).

Fixes #753

### DIFF
--- a/src/Matterhorn/Config.hs
+++ b/src/Matterhorn/Config.hs
@@ -141,7 +141,9 @@ fromIni = do
         threadOrientationField
         (configThreadOrientation defaultConfig)
     configToken <- (Just . TokenCommand  <$> field "tokencmd") <!>
-                  pure Nothing
+                   pure Nothing
+    configOTPToken <- (Just . OTPTokenCommand  <$> field "otptokencmd") <!>
+                      pure Nothing
     configUnsafeUseHTTP <-
       fieldFlagDef "unsafeUseUnauthenticatedConnection" False
     configValidateServerCertificate <-
@@ -374,6 +376,7 @@ defaultConfig =
            , configUrlPath                     = Nothing
            , configPass                        = Nothing
            , configToken                       = Nothing
+           , configOTPToken                    = Nothing
            , configTimeFormat                  = Nothing
            , configDateFormat                  = Nothing
            , configTheme                       = Nothing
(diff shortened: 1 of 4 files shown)


src/Matterhorn/Config.hs
```diff
@@ -375,9 +378,19 @@ getConfig fp = do
                 Just (TokenString _) -> error $ "BUG: getConfig: token in the Config was already a TokenString"
                 Nothing -> return Nothing
 
+            actualOTPToken <- case configOTPToken conf of
+                Just (OTPTokenCommand cmdString) -> do
+                    let (cmd:rest) = T.unpack <$> T.words cmdString
```
jtdaugherty, line 383: This will crash if `cmdString` has less than two words. Can you check for that case, please?

jtdaugherty:
I left one comment but otherwise this looks good to me. Once you've addressed the comment, we can look into merging. There's currently a conflict so you could either rebase on top of `develop` or resolve the conflict. I don't have a preference about which approach you take.

src/Matterhorn/Config.hs
msm-code: Sure. To be honest, I've copied that check from `actualToken` to fix that. Looks like there's a bit of code duplication in the branches for actualPass, actualToken and actualOTPToken - is this going to be a problem?

msm-code:
I prefer to rebase my work - did it now, and (after lengthy cabal update) confirmed that it still works.

src/Matterhorn/Config.hs
jtdaugherty: No, I can clean that up.

jtdaugherty:
Thank you!

PR #803, opened 2023-05-04
Support building with GHC 9.6

RyanGlScott:
For the most part, this simply requires bumping upper version bounds. I did need to make a minor code change to support `mtl-2.3.*`, which no longer re-exports as many things (e.g., `MonadTrans(lift)`) as previous versions do.


jtdaugherty:
Thank you!

PR #839, opened 2024-11-20
notifyV2: shellcheck pass, add appname & icon

fictitiousexistence:
Update echo to printf
Pass shellcheck
Quote unquoted variables
Use $(...) notation instead of legacy backticks
Use foo () instead of function foo
Add appname and icon options for notify-send

### DIFF
--- a/notification-scripts/notifyV2
+++ b/notification-scripts/notifyV2
@@ -12,26 +12,27 @@
 
 # See docs/notification-scripts.md for further details.
 
-PGM="`basename $0`"
-function fail {
-	echo $PGM: "$@" >/dev/stderr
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks for this patch!

For future contributions, something to keep in mind is that these changes need to be committed with logical separation. For example, the change to set the application name and introduce the icon argument are new *features* that are separate from linting/cleanup introduced by some of the other changes. Having them all in one commit makes it more difficult to selectively revert some of them if there's a problem for some reason, and in general it makes commits more difficult to review.

jtdaugherty:
(That sort of thing can usually be done very easily by using `git add -p` rather than `git add`.)

PR #846, opened 2025-04-24
Added notification version 4 with channel name and type

### LINKED ISSUE #805: Add notification version 4 with source channel name and type
I often receive messages from users but I've multiple channels opened with them and I'm not sure from where the message is coming from (private message or a specific channel?).

So it would be nice to provide the notification source information too.

ethernet-zero:
Decided to give it a go on including the channel name and type on the notification payload, so here is notification version 4.

(fixes #805)

### DIFF
--- a/docs/notification-scripts.md
+++ b/docs/notification-scripts.md
@@ -42,7 +42,7 @@ The interface for notification scripts is as follows:
   passing notification details via JSON to the command's standard input.
   The JSON will be UTF-8 encoded; the notification script is responsible
   for decoding it.
-  * NotifyV2 JSON payload fields:
+  * NotifyV3 JSON payload fields:
     * "version": "3"
     * "from": "sender name"
     * "message": "message text..."
--- a/docs/notification-scripts.md
+++ b/docs/notification-scripts.md
@@ -60,6 +60,36 @@ The interface for notification scripts is as follows:
       * `ephemeral`
       * `unknown`
 
+* NotifyV4: Matterhorn will invoke the command with zero arguments,
+  passing notification details via JSON to the command's standard input.
+  The JSON will be UTF-8 encoded; the notification script is responsible
+  for decoding it.
+  * NotifyV4 JSON payload fields:
+    * "version": "4"
+    * "from": "sender name"
+    * "message": "message text..."
+    * "mention": boolean (true or false: did sender mention me?)
+    * "messageType": a string from the following list based on the
+      message type:
+      * `joinChannel`
+      * `leaveChannel`
+      * `addToChannel`
+      * `removeFromChannel`
+      * `headerChange`
+      * `displayNameChange`
+      * `purposeChange`
+      * `channelDeleted`
+      * `ephemeral`
+      * `unknown`
+    * "channel": the channel slug
+    * "channelType": a string from the following list based on the
+      type of channel where the notification originated:
+      * `D` for direct messages
+      * `O` for ordinary channels
+      * `P` for private channels
+      * `G` for group DMs
+      * `U` for unknown types
+
 * Matterhorn will wait for the process to terminate. If the process
   emits any output to standard out OR if the command exits with a
   non-zero exit status, Matterhorn will consider that evidence that
--- a/src/Matterhorn/Config.hs
+++ b/src/Matterhorn/Config.hs
@@ -339,8 +339,9 @@ notifyVersion t =
         "1" -> Right NotifyV1
         "2" -> Right NotifyV2
         "3" -> Right NotifyV3
+        "4" -> Right NotifyV4
         _ -> Left ("Invalid value " <> show t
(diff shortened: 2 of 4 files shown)


docs/notification-scripts.md
```diff
@@ -42,7 +42,7 @@ The interface for notification scripts is as follows:
   passing notification details via JSON to the command's standard input.
   The JSON will be UTF-8 encoded; the notification script is responsible
   for decoding it.
-  * NotifyV2 JSON payload fields:
+  * NotifyV3 JSON payload fields:
```
jtdaugherty, line 45: Thanks for fixing this typo!

src/Matterhorn/Config.hs
```diff
@@ -338,8 +338,10 @@ notifyVersion t =
     case t of
         "1" -> Right NotifyV1
         "2" -> Right NotifyV2
+        "3" -> Right NotifyV3
```
jtdaugherty, line 341: Thanks also for adding this missing line. In the future, please include fixes like this in their own commits separately from commits that add new features. That way, if I needed to revert the feature commit due to a mistake, I would be able to leave these fixes in place.

src/Matterhorn/State/Messages.hs
```diff
@@ -850,6 +873,31 @@ notifyGetPayload NotifyV3 st post mentioned = do
                 PostTypeChannelDeleted    -> "channelDeleted"
                 PostTypeEphemeral         -> "ephemeral"
                 PostTypeUnknown _         -> "unknown"
+notifyGetPayload NotifyV4 st post mentioned = do
+    let notification = NotificationV4 4 msg mentioned sender msgTy chName chType
+    return (A.encode notification)
+        where
+            chan = fromJust $ (findChannelById (post^.postChannelIdL) (st^.csChannels))
```
jtdaugherty, line 880: This `fromJust` will raise a runtime exception if the channel is missing, which could happen due to a race. Instead, please move this into the `Maybe` monad above the `let` binding above, since this function can already fail gracefully with a `Nothing`:
```haskell
chan <- findChannelById (post^.postChannelIdL) (st^.csChannels)
```

src/Matterhorn/State/Messages.hs
```diff
@@ -850,6 +873,31 @@ notifyGetPayload NotifyV3 st post mentioned = do
...
+            chName = chan^.ccInfo.cdName
+            chType = case (chan^.ccInfo.cdType) of
+                Direct      -> "D"
+                Ordinary    -> "O"
+                Private     -> "P"
+                Group       -> "G"
+                (Unknown t) -> "U"
+            msg = sanitizeUserText $ postMessage post
+            sender = maybePostUsername st post
+            msgTy = case postType post of
```
jtdaugherty, line 890: This code appears to duplicate the definition of `msgTy` in the `NotifyV3` case. Would you be willing to factor both (and any other such duplications) out as a top-level function?

jtdaugherty (changes requested):

jtdaugherty:
Thank you for doing this work! I left a few comments in my review. Once we can resolve those, I'll be happy to merge this.

src/Matterhorn/State/Messages.hs
ethernet-zero: Oh, nice! Thanks for the suggestion! :)

src/Matterhorn/State/Messages.hs
ethernet-zero: Sure thing!

src/Matterhorn/State/Messages.hs
```diff
@@ -839,17 +861,39 @@ notifyGetPayload NotifyV3 st post mentioned = do
...
-                PostTypeDisplayNameChange -> "displayNameChange"
-                PostTypePurposeChange     -> "purposeChange"
-                PostTypeChannelDeleted    -> "channelDeleted"
-                PostTypeEphemeral         -> "ephemeral"
-                PostTypeUnknown _         -> "unknown"
+            msgTy = notifyGetMessageType post
+notifyGetPayload NotifyV4 st post mentioned = do
+    chan <- findChannelById (post^.postChannelIdL) (st^.csChannels)
+    let chName = chan^.ccInfo.cdName
+    let chType = notifyGetChannelType chan
```
ethernet-zero, line 868: I'm not 100% sure if this is the right approach, but if I let them in the `where` context then the `chan` variable was out of scope there.

jtdaugherty: Yeah, that ends up being necessary for exactly the reason you found. I would only recommend one minor style change: `let`s can be batched like so:
```haskell
let foo = ...
    bar = ...
```
instead of one `let` per binding.

ethernet-zero: Ah, neat, thanks!

jtdaugherty:
Thanks again for working on this!

PR #848, opened 2025-04-25
Fix the -i/--ignore-configuration option.

### LINKED ISSUE #847: The `ignore-config` option still loads configuration files in default locations
When running Matterhorn with the `-i`/`--ignore-configuration` option, it still loads settings from any configuration file found in default locations. I found this out when trying to generate a release tarball in my machine: the script failed because I had defined some keybindings in my configuration and they didn't match the default ones:

```
%
% Verify that the keybinding document is current
%
31,32c31,32
< | `C-Up` | `channel-list-scroll-up` | Scroll up in the channel list |
< | `C-Down` | `channel-list-scroll-down` | Scroll down in the channel list |
---
> | `M-PgUp` | `channel-list-scroll-up` | Scroll up in the channel list |
> | `M-PgDown` | `channel-list-scroll-down` | Scroll down in the channel list |
35,36c35,36
< | `C-n` | `focus-next-channel` | Change to the next channel in the channel list |
< | `C-p` | `focus-prev-channel` | Change to the previous channel in the channel list |
---
> | `C-Down` | `focus-next-channel` | Change to the next channel in the channel list |
> | `C-Up` | `focus-prev-channel` | Change to the previous channel in the channel list |
85,86c85,86
< | `C-n` | `focus-next-channel` | Select next match |
< | `C-p` | `focus-prev-channel` | Select previous match |
---
> | `C-Down` | `focus-next-channel` | Select next match |
> | `C-Up` | `focus-prev-channel` | Select previous match |
```

My guess would be that the issue stems from [these lines](https://github.com/matterhorn-chat/matterhorn/blob/develop/programs/Main.hs#L22-L23) and more specifically from the fact that `findConfig` still tries to locate the config file in the default locations because when it gets `Nothing` as its argument it can't differentiate if it was an explicit request to not load any configuration files or if no configuration file was explicitly passed using the `-c`/`--config` option.

In short, basically, `optConfLocation opts` can also be `Nothing`, so `findConfig` can receive the same value for two completely different conditions.

ethernet-zero:
Make the code that loads the configuration explicitly return the default configuration without any config file lookups if the option to ignore the configuration is used. The current implementation tries to make the function that looks up the configuration file not load anything by passing a null value to it, but that behavior clashes with the behavior of the `-c`/`--config` option, which when not used results in the same value being passed to the function, so it doesn't have a way of differentiating between the two. As a result, if you have a configuration file in any of the default locations it'll load it regardless.

(Fixes (#847)

### DIFF
--- a/programs/Main.hs
+++ b/programs/Main.hs
@@ -19,9 +19,7 @@ main :: IO ()
 main = do
     opts <- grabOptions
 
-    configResult <- findConfig $ if optIgnoreConfig opts
-                                 then Nothing
-                                 else optConfLocation opts
+    configResult <- tryLoadConfig (optIgnoreConfig opts) (optConfLocation opts)
 
     config <- case configResult of
         Left err -> do
--- a/scripts/local-mkrelease.sh
+++ b/scripts/local-mkrelease.sh
@@ -163,7 +163,7 @@ echo
 echo %
 echo % Verify that the command document is current
 echo %
-diff docs/commands.md <(./run.sh -m -f markdown)
+diff docs/commands.md <(./run.sh -i -m -f markdown)
 echo
 
 TMPDIR=$(mktemp -d)
--- a/src/Matterhorn/Config.hs
+++ b/src/Matterhorn/Config.hs
@@ -5,6 +5,7 @@ module Matterhorn.Config
   ( Config(..)
   , PasswordSource(..)
   , findConfig
+  , tryLoadConfig
   , configConnectionType
   )
 where
@@ -448,13 +449,16 @@ defaultConfig = addDefaultKeys $
            , configCharacterWidths             = Nothing
            }
 
+defaultWarnConfig :: ([String], Config)
+defaultWarnConfig = ([], defaultConfig)
+
 findConfig :: Maybe FilePath -> IO (Either String ([String], Config))
 findConfig mPath = runExceptT $ do
     -- Load the main configuration
     locatedConfig <- lift $ locateConfig configFileName
 
     (warns, config) <- case mPath <|> locatedConfig of
-        Nothing -> return ([], defaultConfig)
+        Nothing -> return defaultWarnConfig
         Just path -> loadConfig path
 
     config' <- fixupPaths config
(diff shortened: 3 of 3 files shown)


programs/Main.hs
```diff
@@ -19,9 +19,7 @@ main :: IO ()
 main = do
     opts <- grabOptions
 
-    configResult <- findConfig $ if optIgnoreConfig opts
-                                 then Nothing
-                                 else optConfLocation opts
+    configResult <- tryLoadConfig (optIgnoreConfig opts) (optConfLocation opts)
```
jtdaugherty, line 22: Thank you so much for working on this! Looking over the implementation, I'm thinking that rather than pushing the consideration of the config setting down into a thin wrapper around `findConfig`, I'd prefer it if the caller took care of that (i.e., `main`). My reasoning is that the decision about whether to use the default config is unrelated to loading it once we've decided not to use the default, so I hesitate to bundle those decisions together and call them `tryLoadConfig`. I am happy to make the change if you like, but if you are willing, I think it would be something along these lines:
* Remove `tryLoadConfig`,
* Export `defaultConfig` from `Matterhorn.Config`, and
* In `main`, `case` on `optIgnoreConfig opts` and just go ahead and return the default configuration then and there.

One other thought: I really appreciate that the commit message includes some details about what this is doing. For fixes like this, it would be additionally helpful if the commit message said some things about what the old (undesirable) behavior was and how this fixes that. I see that this links to a ticket and I'm glad that's there, but repos can also become detached from their hosting services (or a developer who's offline can't access them) in which case having the context in the commit message itself is extremely helpful. Here's an example of the sort of exposition I'm thinking of:

https://github.com/matterhorn-chat/matterhorn/commit/a717900781cb5081f6eaf423eecd97a63a0874ff

jtdaugherty (changes requested):

programs/Main.hs
ethernet-zero: Actually, exporting `defaultConfig` was my first thought, but I figured that maybe it was preferable to have everything related to loading configuration in the Config module. But yeah, I can change that, no problem.

I'll also keep in mind your suggestion about the commit messages. Thanks a lot!

jtdaugherty: > maybe it was preferable to have everything related to loading configuration in the Config module

Yes, this would also be my preference! I think by keeping `defaultConfig` in that module, we satisfy this goal. The thing that makes this change different is that it is a higher-level application policy choice about where to get the config from.

ethernet-zero: Good point. :smiley:

jtdaugherty: Looking at it a bit, I'm thinking that `defaultWarnConfig` might be a slightly strange thing to export because it bundles an empty warning list, and that is a policy choice that is separate from the default configuration itself. The emptiness of the warning list is a detail that is specific only to `findConfig`, and if we remove `tryLoadConfig` as I mentioned above, that leaves only `defaultConfig` as the thing that needs to be exported. I'm happy to make those changes, though, so if it would be more expedient I am happy to merge this and then make those adjustments.

jtdaugherty:
Thanks for working on this!

